PR #7, opened 2020-12-07
Added Exoscale support

Xartos:
Updated kubespray to 2.14.2 (to support ubuntu 20.04) and pushed exoscale support

### DIFF
--- a/terraform/esys-exoscale/README.md
+++ b/terraform/esys-exoscale/README.md
@@ -1,0 +1,29 @@
+# Terraform for Exoscale kubespray cluster
+
+## Quickstart
+
+Edit tfvars to match your setup
+
+```
+$EDITOR tfvars.json
+```
+
+Create a encrypted file with your Exoscale credentials
+
+```
+cat << EOF > secrets.env
+TF_VAR_exoscale_api_key=
+TF_VAR_exoscale_secret_key=
+EOF
+sops --encrypt --in-place --pgp <PGP key fingerprint> secrets.env
+sops secrets.env
+```
+
+Run terraform and create your inventory
+
+```
+sops exec-env secrets.env 'terraform apply -var-file tfvars.json'
+./generate-inventory.sh terraform.tfstate
+mv sc-inventory.ini $CK8S_CONFIG_PATH/sc-config/inventory.ini
+mv wc-inventory.ini $CK8S_CONFIG_PATH/wc-config/inventory.ini
+```
--- a/terraform/exoscale/modules/kubernetes-cluster/templates/worker-cloud-init.tmpl
+++ b/terraform/exoscale/modules/kubernetes-cluster/templates/worker-cloud-init.tmpl
@@ -1,0 +1,26 @@
+#cloud-config
+%{ if ceph_enabled }
+bootcmd:
+- [ cloud-init-per, once, move-second-header, sgdisk, --move-second-header, /dev/vda ]
+- [ cloud-init-per, once, create-ceph-part, parted, --script, /dev/vda, 'mkpart 2 ${ceph_partition_size}GB -1' ]
+%{ endif }
+
+write_files:
+  - path: /etc/netplan/eth1.yaml
+    content: |
+      network:
+        version: 2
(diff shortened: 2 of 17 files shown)


terraform/esys-exoscale/README.md
```diff
@@ -0,0 +1,29 @@
...
+Create a encrypted file with your Exoscale credentials
+
+```
+cat << EOF > secrets.env
+TF_VAR_exoscale_api_key=
+TF_VAR_exoscale_secret_key=
+EOF
+sops --encrypt --in-place --pgp <PGP key fingerprint> secrets.env
+sops secrets.env
+```
```
cristiklein, line 20: I would rather use `.cloudstack.ini` or environment variables.

cristiklein (commented):
I think we should upstream the Exoscale Terraform part?

Also, I would rather we "force" users to provide their credentials with `~/.cloudstack.ini` or environment variables, instead of luring them into committing them to a configuration file.

Finally, we are slowly moving away from having a pair of SC-WC clusters, but I guess that is a discussion for another day. :smile:

Xartos:
> I think we should upstream the Exoscale Terraform part?
> 
> Also, I would rather we "force" users to provide their credentials with `~/.cloudstack.ini` or environment variables, instead of luring them into committing them to a configuration file.
> 
> Finally, we are slowly moving away from having a pair of SC-WC clusters, but I guess that is a discussion for another day. smile

I thought so too, but I was not allowed to allocate time for upstreaming. The same answer goes to the next point as well, I tried to allocate as little time as possible for this so I didn't test how and if that works. But I agree that that would be far nicer!

That's why I separated the two, so that you can set up clusters one by one, but at the same time we have an easy way of setting up a cluster pair that works with apps.

cristiklein:
@Xartos One more. :smile: 

For SSH key management, I just tested and we can choose *NOT* to create an `ssh_key` resource, leave the `ssh_key` attribute of the virtual machines empty and use `cloud-init`'s `ssh_authorized_keys` attribute to configure instance SSH keys. I see two advantage with the `cloud-init` approach:
1. More portable across cloud providers.
2. Easier to add several SSH keys, e.g., if the cluster is managed by a team, but each team member is identified by their own SSH key.
3. (Bonus) One Terraform resource less. :smile: 

What do you think?

terraform/exoscale/modules/kubernetes-cluster/templates/worker-cloud-init.tmpl
```diff
@@ -0,0 +1,26 @@
+#cloud-config
+%{ if ceph_enabled }
+bootcmd:
+- [ cloud-init-per, once, move-second-header, sgdisk, --move-second-header, /dev/vda ]
+- [ cloud-init-per, once, create-ceph-part, parted, --script, /dev/vda, 'mkpart 2 ${ceph_partition_size}GB -1' ]
```
danielharr, line 5: `${ceph_partition_size}GB` here is actually how much space to leave before creating the ceph partition, so the variable name becomes misleading. The ceph partition will start at `${ceph_partition_size}GB` and grow until the next partition starts or end of disk :confused:

cristiklein: I am happy to take the blame for this confusion. Should be called `rootfs_partition_size` or `non_ceph_partition_size` or `ceph_partition_start`.

Xartos: I'm a bit split to how we should approach this for the upstream version. Should we include this as it is right now *(with the variable name change)* or remove it from the upstreamed version and only have it for the "elastisys" version? Since it kind of forces people

cristiklein: I would argue that the `ceph_enabled` flag (I assume it's disabled by default) makes it un-opinionated. I would push it upstream.

cristiklein:
This is super-done. :smile:

PR #8, opened 2020-12-10
Initial rook support

danielharr:
This adds initial support for installing Rook on a fresh Kubernetes installation.

Only RBD volumes has been tested. The CephFS driver has been disabled to save resources.

So far monitoring is disabled by default, since it is not assumed that the Prometheus Operator CRDs are available.
Monitoring can be enabled at a later stage by setting `spec.monitoring.enabled=true` in `cluster.yaml` and reapplying.

The resource requests are an initial guess from running a Compliant Kubernetes service cluster with Rook, and does therefore not necessarily make sense.

For testing purposes, it could make sense to
* reduce the `mon` replica count to 1
* reduce the storageclass replication to 1

### DIFF
--- a/rook/README.md
+++ b/rook/README.md
@@ -1,0 +1,46 @@
+# Rook installation
+
+Also see the official documentation at <https://rook.io/docs/rook/v1.5/> and the repository at <https://github.com/rook/rook>.
+
+## Quickstart
+
+```bash
+# Make sure your KUBECONFIG and Kubernetes context are set correctly
+./deploy-rook.sh
+```
+
+## Requirements
+
+### Ceph partition or disk
+
+An empty disk or partition needs to be available.
+Ceph will create an [OSD](https://docs.ceph.com/en/latest/man/8/ceph-osd/) for each partition or disk to provide access to it.
+
+To create an empty partition `/dev/vda2` on the `/dev/vda` disk, leaving 50 GB to other partitions, use the following cloud-init config:
+
+```yaml
+#cloud-config
+bootcmd:
+- [ cloud-init-per, once, move-second-header, sgdisk, --move-second-header, /dev/vda ]
+- [ cloud-init-per, once, create-ceph-part, parted, --script, /dev/vda, 'mkpart 2 50GB -1' ]
+```
+
+### Nodes for Ceph mons
+
+The Ceph [`mon`s](https://docs.ceph.com/en/latest/man/8/ceph-mon/) should not run on the same node.
+Make sure that there are at least as many worker nodes as there are replicas of `mon`s.
+See `spec.mon.count` in [cluster.yaml](./cluster.yaml).
+
+### Monitoring with Prometheus
+
+Prometheus Operator CRDs should be installed before applying the Ceph cluster manifest if `spec.monitoring.enabled=true`.
+
+## Installation
+
+See [deploy-rook.sh](./deploy-rook.sh) for an example.
+
+## Issues
+
+If the operator seems to be unable to remove configmaps or deployments, make sure that the Kubernetes API server is able to process requests.
+For example, if cert-manager custom resources are created before cert-manager is installed, we experienced cases where the Kubernetes API server did not process requests from the Rook operator.
+This was resolved once cert-manager was installed.
--- a/rook/deploy-rook.sh
+++ b/rook/deploy-rook.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+helm repo add rook-release https://charts.rook.io/release
+
+namespace="rook-ceph"
+release_name="rook-ceph"
+chart="rook-release/rook-ceph"
+chart_version="v1.5.3"
+
+# Install rook operator
+kubectl create namespace $namespace --dry-run -o yaml | kubectl apply -f -
+helm upgrade --install --namespace $namespace $release_name $chart \
+  --version $chart_version --values operator-values.yaml --wait
+
+# Install ceph cluster
+kubectl --namespace $namespace apply -f cluster.yaml
+
+# Create storageclass
+kubectl --namespace $namespace apply -f storageclass.yaml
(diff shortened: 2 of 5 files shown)


rook/README.md
```diff
@@ -0,0 +1,39 @@
...
+#cloud-config
+bootcmd:
+- [ cloud-init-per, once, move-second-header, sgdisk, --move-second-header, /dev/vda ]
+- [ cloud-init-per, once, create-ceph-part, parted, --script, /dev/vda, 'mkpart 2 50GB -1' ]
+```
+
+### Nodes for Ceph mons
+
+The Ceph [`mon`s](https://docs.ceph.com/en/latest/man/8/ceph-mon/) should not run on the same node.
+Make sure that there are at least as many nodes as there are replicas of `mon`s.
```
Xartos, line 30:
```suggestion
Make sure that there are at least as many worker nodes as there are replicas of `mon`s.
```

rook/deploy-rook.sh
```diff
@@ -0,0 +1,21 @@
...
+
+helm repo add rook-release https://charts.rook.io/release
+
+namespace="rook-ceph"
+release_name="rook-ceph"
+chart="rook-release/rook-ceph"
+chart_version="v1.5.3"
+
+# Install rook operator
+kubectl create namespace $namespace
```
Xartos, line 13: I actually think we should have this as a file and use `apply -f` since we cannot run this multiple times atm.

rook/deploy-rook.sh
```diff
@@ -0,0 +1,21 @@
...
+helm install --namespace $namespace $release_name $chart \
```
Xartos, line 14: In the same spirit we should change this to
```suggestion
helm update --install --namespace $namespace $release_name $chart \
```

rook/deploy-rook.sh
```diff
@@ -0,0 +1,21 @@
...
+  --version $chart_version --values operator-values.yaml --wait
+
+# Install ceph cluster
+kubectl --namespace $namespace create -f cluster.yaml
```
Xartos, line 18:
```suggestion
kubectl --namespace $namespace apply -f cluster.yaml
```

rook/deploy-rook.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Create storageclass
+kubectl --namespace $namespace create -f storageclass.yaml
```
Xartos, line 21:
```suggestion
kubectl --namespace $namespace apply -f storageclass.yaml
```

rook/storageclass.yaml
```diff
@@ -0,0 +1,33 @@
+# See https://rook.io/docs/rook/v1.5/ceph-pool-crd.html
+apiVersion: ceph.rook.io/v1
+kind: CephBlockPool
+metadata:
+  name: replicapool
+  namespace: rook-ceph
+spec:
+  failureDomain: host
+  replicated:
+    size: 2
```
Xartos, line 10: This must also match be maximum the nr of worker nodes right? Maybe add it to the README

Xartos (changes requested):
One minor note on idempotency

viktor-f (commented):
You could perhaps also make some note of the problems we had with cert-manager. It isn't really connected to rook, but others might also stumble into the same issue.

rook/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+## Requirements
+
+### Ceph partition or disk
+
+An empty disk or partition needs to be available.
+Ceph will create an [OSD](https://docs.ceph.com/en/latest/man/8/ceph-osd/) for each partition or disk to provide access to it.
+
+To create an empty partition `/dev/vda2` on the `/dev/vda` disk, leaving 50 gb to other partitions, use the following cloud-init config:
```
cristiklein, line 18: Nit-pick: Please capitalize "GB" (and teach me how to make suggestions on GitHub). :smile:

rook/README.md
```diff
@@ -0,0 +1,39 @@
+# Rook installation
+
+Also see the official documentation at <https://rook.io/docs/rook/v1.5/> and the repository at <https://github.com/rook/rook>.
+
+## Quickstart
+
+```bash
+KUBECONFIG=/path/to/kubeconfig ./deploy-rook.sh
```
cristiklein, line 8: Can we do:
```
# Make sure your KUBECONFIG and Kubernetes context are set correctly
./deploy-rook.sh
```

rook/deploy-rook.sh
cristiklein: :+1:

rook/deploy-rook.sh
cristiklein: :+1:

cristiklein (changes requested):

rook/README.md
viktor-f: When you create a comment in github it's the first button to the right of `write` and `preview`.

rook/storageclass.yaml
danielharr: No, I think this is just how many times ceph should replicate the data on volumes with this storage class.
Maybe I'm wrong, but I've been running like this anyway so it's not a "must" at least :slightly_smiling_face:

rook/README.md
danielharr: Good point :slightly_smiling_face:

rook/deploy-rook.sh
danielharr: Sure, changed it to `kubectl create | apply` though since it's a plain namespace object :slightly_smiling_face:

danielharr:
Addressed your comments, PTAL and I'll add more commits or fixup for merge :slightly_smiling_face:

cristiklein (approved):

rook/storageclass.yaml
Xartos: Yea, I might be completely wrong here. I just remembered fiddling around with some of these numbers to get one cluster setup to work. But it must've been the mons then

Xartos (approved):
LGTM!

viktor-f (approved):

PR #9, opened 2020-12-14
Updated to kubespray v2.14.2

Xartos:
To support Ubuntu 20.04 *(and just to update to the latest version as soon as possible)* we want to update kubespray to the latest version. It works with the same inventory as v2.13

### DIFF
--- a/config/group_vars/k8s-cluster/k8s-cluster.yml
+++ b/config/group_vars/k8s-cluster/k8s-cluster.yml
@@ -1,3 +1,4 @@
+---
 # Kubernetes configuration dirs and system namespace.
 # Those are where all the additional config stuff goes
 # the kubernetes normally puts in /srv/kubernetes.
@@ -19,7 +20,7 @@ kube_users_dir: "{{ kube_config_dir }}/users"
 kube_api_anonymous_auth: true
 
 ## Change this to use another Kubernetes version, e.g. a current beta release
-kube_version: v1.17.12
+kube_version: v1.18.10
 
 # kubernetes image repo define
 kube_image_repo: "k8s.gcr.io"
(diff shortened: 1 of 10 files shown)


lentzi90 (approved):
LGTM!
Just one note: even though it works with the same inventory, I guess there may be some changes in the defaults? Probably not a big deal but I think we should check how much we differ from the defaults every now and then so we don't run with old settings "just because".

viktor-f (approved):

OlleLarsson (approved):

Xartos:
> LGTM!
> Just one note: even though it works with the same inventory, I guess there may be some changes in the defaults? Probably not a big deal but I think we should check how much we differ from the defaults every now and then so we don't run with old settings "just because".

Good idea!
I've copied the default inventory from `kubespray/inventory/sample/group_vars` to `config/group_vars/` but kept the changes that we need.
You can run `diff --color -r  kubespray/inventory/sample/group_vars/ config/group_vars/` to see what we need to differ. But this should be a task everytime we update kubespray I think! I don't think it's so hard, and it's good to see what variables that differs for us and maybe we can move towards kubespray defaults more and more

viktor-f (approved):

config/group_vars/k8s-cluster/k8s-cluster.yml
```diff
@@ -77,12 +87,12 @@ kube_network_plugin: calico
...
 
 # Kubernetes internal network for services, unused block of space.
-kube_service_addresses: 10.96.0.0/12
+kube_service_addresses: 10.233.0.0/18
 
 # internal network. When used, it will assign IP
 # addresses from this range to individual pods.
 # This network must be unused in your network infrastructure!
-kube_pods_subnet: 192.168.0.0/16
+kube_pods_subnet: 10.233.64.0/18
```
lentzi90, line 95: It probably doesn't matter much but these are now different from what we have in ck8s-cluster. In any case I guess any terraform code written to work with kubespray will probably have defaults that work with the default kubespray values rather than default ck8s-cluster values so I think this change is fine. :slightly_smiling_face:

lentzi90 (approved):

config/group_vars/k8s-cluster/k8s-cluster.yml
Xartos: Yea, I thought so as well. And these ranges shouldn't matter for anyone anyway since we should use DNS for everything anyway. So it might be a good test to see if apps works with this :)

PR #10, opened 2020-12-14
bin: ansible options can be passed via wrapper

### LINKED ISSUE #6: [1] Support ask become password
Some environments do not have passwordless sudo, which means that we need a way to tell ansible that it should ask for the password. We could for example allow passing extra flags through the wrapper to ansible so that we can set `--ask-become-pass`.

viktor-f:
This PR adds the option to pass arguments to ansible through the wrapper.

fixes #6 

Examples: 
```
./bin/ck8s-kubespray apply wc --version
...
[ck8s] Running kubespray
ansible-playbook 2.9.6
...
```

```
./bin/ck8s-kubespray run-playbook wc facts.yml --help
...
[ck8s] Running kubespray
usage: ansible-playbook [-h] [--version] [-v] [-k]
                        [--private-key PRIVATE_KEY_FILE] [-u REMOTE_USER]
                        [-c CONNECTION] [-T TIMEOUT]
                        [--ssh-common-args SSH_COMMON_ARGS]
                        [--sftp-extra-args SFTP_EXTRA_ARGS]
                        [--scp-extra-args SCP_EXTRA_ARGS]
...
```

### DIFF
--- a/bin/apply.bash
+++ b/bin/apply.bash
@@ -19,14 +19,16 @@ source venv/bin/activate
 pip install -r requirements.txt
 
 log_info "Running kubespray"
-sops_exec_file_no_fifo "${secrets[ssh_key]}" "ansible-playbook -i "${config[inventory_file]}" cluster.yml -b --private-key {}"
+sops_exec_file_no_fifo "${secrets[ssh_key]}" "ansible-playbook -i "${config[inventory_file]}" cluster.yml -b --private-key {} ${@}"
 
 popd
 
 log_info "Kubespray done"
 
-mv "${config_path}/artifacts/admin.conf" "${secrets[kube_config]}"
-sops_encrypt "${secrets[kube_config]}"
+if [ -f "${config_path}/artifacts/admin.conf" ]; then
+    mv "${config_path}/artifacts/admin.conf" "${secrets[kube_config]}"
+    sops_encrypt "${secrets[kube_config]}"
+fi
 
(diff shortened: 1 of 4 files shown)


lentzi90 (approved):
LGTM!

bin/apply.bash
```diff
@@ -19,14 +19,16 @@ source venv/bin/activate
 pip install -r requirements.txt
 
 log_info "Running kubespray"
-sops_exec_file_no_fifo "${secrets[ssh_key]}" "ansible-playbook -i "${config[inventory_file]}" cluster.yml -b --private-key {}"
+sops_exec_file_no_fifo "${secrets[ssh_key]}" "ansible-playbook -i "${config[inventory_file]}" cluster.yml -b --private-key {} ${@}"
```
Xartos, line 22: Would it be possible for this script to only run `./run-playbook.bash cluster.yml ${@}`?
Not part of this issue, but just a thought that popped up.

Xartos (approved):
LGTM!

bin/apply.bash
viktor-f: Hmm, yeah that should work. The only difference should be some of the terminal messages

Xartos: I think it would be nice. But if you don't want to do it here we can always add another PR for this

viktor-f: Sounds good. I'll merge this now then.

PR #22, opened 2021-01-22
Simplified the init flow and made it a bit more transparent

### LINKED ISSUE #17: [1] Change init flow to increase transparency in how  kubespray is used
Instead of copying the whole `group_vars` folder, we should focus on only setting those variables whose value we need to change. This will not only improve adherence to the Principle of Least Astonishment (POLA), but also improve ease of maintenance moving forwards.

Relevent Ansible documentation on how variables get merged is [here](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#how-variables-are-merged).

Acceptance criteria:
* This repo only contains kubespray variables we need to change.
* Configuration files generated by `./bin/ck8s-kubespray init` only contains variables we need to change.
* `./bin/ck8s-kubespray apply` consumes an intuitive merge of kubespray defaults and the variables we overwrite.
* Update documentation (see https://github.com/elastisys/compliantkubernetes/issues/44)

### LINKED ISSUE #18: [2] Update to Kubespray 2.15
This would fix a few errors, such as:

```
Warning: Interpolation-only expressions are deprecated
 
  on create-infrastructure.tf line 66, in resource "aws_instance" "bastion-server":
  66:     "Cluster", "${var.aws_cluster_name}",
 
Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.
 
Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.
 
 
Warning: Quoted type constraints are deprecated
 
  on variables.tf line 47, in variable "aws_cidr_subnets_private":
  47:   type        = "list"
 
Terraform 0.11 and earlier required type constraints to be given in quotes,
but that form is now deprecated and will be removed in a future version of
Terraform. To silence this warning, remove the quotes around "list" and write
list(string) instead to explicitly indicate that the list elements are
strings.
 
(and 9 more similar warnings elsewhere)
```

Acceptance criteria:
* compliantkubernetes-kubespray submodules kubespray 2.15.
* This configuration is tested at least with AWS and Exoscale.

Xartos:
Moved the flavors from full configs to diff from default.
This way it's easier to know what the differences are from vanilla kubespray

fixes #17
fixes #18


lentzi90 (commented):
I like the simplification, but I'm not a fan of the "scripts that write yaml". Do we really need to use `update_config` to write out everything like this? It seems to me like we could replace at least the `common.bash` with just a few static files that we copy to the config folder. What do you think?

Edit: I'm basically asking if we could keep the files that were deleted, but reduce them to just the diff like you have done in the scripts now.

Also, do we need the `envsubst` for `inventory.ini`? I mean the `${PREFIX}` is just internal to ansible, right? Or will this show up in Kubernetes also?

config/common.bash
```diff
@@ -0,0 +1,107 @@
...
+admission_plugins_list=( 'PodSecurityPolicy' \
+  'NamespaceLifecycle' \
+  'LimitRanger' \
+  'ServiceAccount' \
+  'DefaultStorageClass' \
+  'DefaultTolerationSeconds' \
+  'MutatingAdmissionWebhook' \
+  'ValidatingAdmissionWebhook' \
+  'ResourceQuota' \
+  'NodeRestriction' )
```
OlleLarsson, line 28:
```suggestion
admission_plugins_list=(
  'PodSecurityPolicy'
  'NamespaceLifecycle'
  'LimitRanger'
  'ServiceAccount'
  'DefaultStorageClass'
  'DefaultTolerationSeconds'
  'MutatingAdmissionWebhook'
  'ValidatingAdmissionWebhook'
  'ResourceQuota'
  'NodeRestriction'
)

```

config/common.bash
```diff
@@ -0,0 +1,107 @@
...
+update_config "${config_path}/group_vars/k8s-cluster/k8s-cluster.yml" 'kube_kubeadm_apiserver_extra_args.delete-collection-workers' '3'
+update_config "${config_path}/group_vars/k8s-cluster/k8s-cluster.yml" 'kube_proxy_mode' 'iptables'
+update_config "${config_path}/group_vars/k8s-cluster/k8s-cluster.yml" 'kube_proxy_metrics_bind_address' '0.0.0.0:10249'
+update_config "${config_path}/group_vars/k8s-cluster/k8s-cluster.yml" 'enable_nodelocaldns' 'false'
+update_config "${config_path}/group_vars/k8s-cluster/k8s-cluster.yml" 'kubeadm_control_plane' 'true'
+update_config "${config_path}/group_vars/k8s-cluster/k8s-cluster.yml" 'kubernetes_audit' 'true'
+update_config "${config_path}/group_vars/k8s-cluster/k8s-cluster.yml" 'audit_log_path' '/var/log/kube-audit/kube-apiserver.log'
+update_config "${config_path}/group_vars/k8s-cluster/k8s-cluster.yml" 'audit_log_maxage' '7'
+
+audit_policy_rules='# Log pod changes at RequestResponse level
```
OlleLarsson, line 39: Just a thought. Should we perhaps have the policy in a file and load it here?

Xartos:
@lentzi90 We still need to merge the files with kubespray config though? Or can you have a key defined multiple times?
And if we need to merge them, I see this as a more transparent solution.
I think kubespray set's that as the k8s name, might be wrong there though.

lentzi90:
I don't understand the merging, maybe I'm missing something. :thinking: Aren't most of the values completely static, so we could merge everything already now and put the resulting files as "examples" that we just copy to the config folder when initializing?

Hmm maybe I misunderstood this from the beginning. Is the merging we do on init "our values" + "kubespray *sample inventory*"? Then I guess the problem is that the sample inventory has some values that are not kubespray defaults so we would need both those values and our own and that is what we try to minimize with this script, correct?
Could we instead remove all kubespray defaults from the sample inventory to make it smaller?

I think there is value in having "our values" visible in plain yml files even though it may require a bit more work to maintain. We could even keep the scripts for generating them to get the best of both.

Xartos:
Isn't the sample inventory the defaults? I thought that we **needed** to use the sample inventory values.
But if we don't need the sample inventory then maybe we can completely skip it and (as you're saying) just have config with "our config".
Do you know any good way of getting the kubespray defaults?

cristiklein:
Is this what you are looking for: https://github.com/kubernetes-sigs/kubespray/blob/master/roles/kubespray-defaults/defaults/main.yaml ?

Xartos:
Right, yes.
But I'm very confused why the sample inventory looks the way it does. All the variables that aren't commented out seems to be the same as the default values (although not all defaults are present). I understand that they want to have some sample to show. But couldn't they just had only commented variables that you could de-comment if you wanted to differ from the default? :dizzy_face: 

Anyway that's a kubespray issue and not ours. Then I guess we should make it as lennart says, and only have the config that differs from the **default** config and not the sample config. And copy that as our config, with no merging/overriding of yaml files.

cristiklein:
@Xartos Could you ask on the `kubespray-dev` Slack? Maybe this is something they would eventually like to have fixed and would be happy to receive a PR.

lentzi90 (approved):
Nice and simple!
Will need a squash before merge, but I guess we wait for some more opinions first. :slightly_smiling_face:

cristiklein:
I love it! So transparent!

@Xartos Can you test if [these instructions](https://compliantkubernetes.io/operator-manual/aws/) still work?

Xartos:
@cristiklein Just to get rid of the ssh keys now :wink: 

Yes, I'll check it.

cristiklein:
@Xartos Should we also aim for the "one script does one thing" approach? I'm drawing inspiration from EKS Distro: https://github.com/aws/eks-distro/blob/main/development/kops/create_configuration.sh.

Or is this another issue/PR?

config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - "ServiceAccount"
+  - "DefaultStorageClass"
+  - "DefaultTolerationSeconds"
+  - "MutatingAdmissionWebhook"
+  - "ValidatingAdmissionWebhook"
+  - "ResourceQuota"
+  - "NodeRestriction"
+
+kube_kubeadm_apiserver_extra_args:
+  delete-collection-workers: "3"
```
OlleLarsson, line 14: Anyone knows the reason for this value?
This feels like a value that we've kept all the way from when we were using rke https://rancher.com/docs/rke/latest/en/example-yamls/

config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+kube_proxy_mode: iptables
+kube_proxy_metrics_bind_address: "0.0.0.0:10249"
+
+enable_nodelocaldns: false
+
+kubernetes_audit: true
+audit_log_path: "/var/log/kube-audit/kube-apiserver.log"
```
OlleLarsson, line 22: Should we perhaps use the default value `/var/log/audit/kube-apiserver-audit.log`?
I know it will require changes to fluentd but maybe it is worth it to be as vanilla as possible.

config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
```diff
@@ -0,0 +1,89 @@
...
+audit_log_maxage: 7
+
```
OlleLarsson, line 24: Should we set `audit_log_maxsize`? Default is 100mb
https://github.com/kubernetes-sigs/kubespray/blob/v2.14.2/roles/kubernetes/master/defaults/main/main.yml#L52
but if i'm not missremebering we generate 1gb+(?) of audit logs every day (i might be confusing this with apiserver logs, not sure).

config/inventory.ini
```diff
@@ -2,26 +2,31 @@
...
 [all]
-${PREFIX}-node1 ansible_user=ubuntu ansible_host=95.54.0.12 etcd_member_name=etcd1 # Master 1
-${PREFIX}-node2 ansible_user=ubuntu ansible_host=95.54.0.13
-${PREFIX}-node3 ansible_user=ubuntu ansible_host=95.54.0.14
+${PREFIX}-node1 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip> etcd_member_name=etcd1 # Master 1
+${PREFIX}-node2 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip> etcd_member_name=etcd2 # Master 2
+${PREFIX}-node3 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip> etcd_member_name=etcd3 # Master 3
+${PREFIX}-node4 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip>
+${PREFIX}-node5 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip>
+r{PREFIX}-node6 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip>
```
OlleLarsson, line 10:
```suggestion
${PREFIX}-node6 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip>
```

config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
Xartos: Most probably, I just copied whatever that we're using from ck8s-cluster repo.
But I can remove it.

config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
Xartos: According to this https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
maxsize is only related to rotation size

```text
The maximum size in megabytes of the audit log file before it gets rotated.
```

So I guess it's okey for it to rotate multiple times a day?

OlleLarsson: Yep :)

config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
OlleLarsson: I'm not sure if we have to remove it.
> Number of workers spawned for DeleteCollection call. These are used to speed up namespace cleanup.

It sounds useful to have more than 1.
I'm fine with keeping it the way it is right now.

Xartos:
@cristiklein I've tested the AWS tutorial, and created this https://github.com/elastisys/compliantkubernetes/pull/47 and updated this repo to make it work properly

config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
Xartos: I tried to see if I could find any discussion about this, but I can't find anything other than its 1 as default and rke sets this to 3.
So I guess there's no big pros/cons for this. So lets leave it for now

config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
```diff
@@ -0,0 +1,94 @@
...
+kube_kubeadm_apiserver_extra_args:
+  delete-collection-workers: "3"
+
+kube_proxy_mode: iptables
+kube_proxy_metrics_bind_address: "0.0.0.0:10249"
+
+enable_nodelocaldns: false
+
+kubernetes_audit: true
+audit_log_path: "/var/log/audit/kube-apiserver-audit.log"
```
OlleLarsson, line 27: Could you open an issue in the apps repo so that we remember to update fluentd to look for this file?

OlleLarsson (approved):

config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
Xartos: https://github.com/elastisys/compliantkubernetes-apps/pull/211

config/aws/group_vars/k8s-cluster/ck8s-k8s-cluster-aws.yaml
```diff
@@ -0,0 +1 @@
+etcd_kubeadm_enabled: false
```
lentzi90, line 1: Is this supposed to be `false` when we have it set to `true` in the default and gcp flavors? Seems very dangerous to me to make them different! [This issue](https://github.com/kubernetes-sigs/kubespray/issues/6687) also seem to indicate that kubeadm managed etcd will be the default in the future.

config/inventory.ini
```diff
@@ -2,26 +2,31 @@
...
 [all]
-${PREFIX}-node1 ansible_user=ubuntu ansible_host=95.54.0.12 etcd_member_name=etcd1 # Master 1
-${PREFIX}-node2 ansible_user=ubuntu ansible_host=95.54.0.13
-${PREFIX}-node3 ansible_user=ubuntu ansible_host=95.54.0.14
+${PREFIX}-node1 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip> etcd_member_name=etcd1 # Master 1
+${PREFIX}-node2 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip> etcd_member_name=etcd2 # Master 2
+${PREFIX}-node3 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip> etcd_member_name=etcd3 # Master 3
+${PREFIX}-node4 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip>
+${PREFIX}-node5 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip>
+${PREFIX}-node6 ansible_user=ubuntu ansible_host=<public ip> ip=<private ip>
```
lentzi90, line 10: Could you also add `access_ip`, or at least a comment with [this](https://kubespray.io/#/docs/vars?id=addressing-variables):
> access_ip - IP for other hosts to use to connect to. Often required when deploying from a cloud, such as OpenStack or GCE and you have separate public/floating and private IPs.

lentzi90 (commented):
A couple of comments below.
I also noticed that we don't set `kubeadm_control_plane: true` anymore. This variable has been removed in kubespray v2.15 (no longer possible to set to `false`), but in v2.14 that is used here it still defaults to `false`. I would suggest either updating the submodule or adding it back.

config/aws/group_vars/k8s-cluster/ck8s-k8s-cluster-aws.yaml
Xartos: This is based on the AWS guide on compliantkubernetes.io and in turn the AWS terraform script in kubespray. Which creates separate etcd nodes. So either we update the terraform scripts or leave it like this.

lentzi90: Ok, then I guess it will have to be this way at least for this PR. I'm uncomfortable having different settings for this in production environments but we are not quite there yet for AWS, right?

Xartos: Yea, I wouldn't say it is

Xartos:
> A couple of comments below.
> I also noticed that we don't set `kubeadm_control_plane: true` anymore. This variable has been removed in kubespray v2.15 (no longer possible to set to `false`), but in v2.14 that is used here it still defaults to `false`. I would suggest either updating the submodule or adding it back.

Right, I thought 2.15 already was merged. I'll update it, or do we see any reason not to?

lentzi90 (approved):
LGTM!

PR #34, opened 2021-01-28
Fixed templates not working when complying with mdl rule MD041

Xartos:
Hard to test these things when it only applies once in main


lentzi90 (approved):
LGTM!

OlleLarsson (approved):

PR #36, opened 2021-02-01
Added some documentation on how to set up on citycloud + flavor

Xartos:
Added openstack/citycloud flavor and added some snippets I used to set it up with.

### DIFF
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -32,7 +32,6 @@ fi
 # Validate the flavor
 if [ "${flavor}" != "default" ] && \
    [ "${flavor}" != "gcp" ] && \
-   [ "${flavor}" != "citycloud" ] && \
    [ "${flavor}" != "openstack" ] && \
    [ "${flavor}" != "aws" ]; then
     log_error "ERROR: Unsupported flavor: ${flavor}"
@@ -84,8 +83,6 @@ elif [[ "${flavor}" == "gcp" ]]; then
 elif [[ "${flavor}" == "aws" ]]; then
   cp -r "${config_defaults_path}/aws/group_vars" "${config_path}/"
 elif [[ "${flavor}" == "openstack" ]]; then
-  cp -r "${config_defaults_path}/openstack/group_vars" "${config_path}/"
-elif [[ "${flavor}" == "citycloud" ]]; then
   cp -r "${config_defaults_path}/openstack/group_vars" "${config_path}/"
 fi
 
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -30,7 +30,11 @@ if [[ ${CK8S_CLOUD_PROVIDER} != "" ]]; then
 fi
 
 # Validate the flavor
-if [ "${flavor}" != "default" ] && [ "${flavor}" != "gcp" ] && [ "${flavor}" != "aws" ]; then
+if [ "${flavor}" != "default" ] && \
+   [ "${flavor}" != "gcp" ] && \
+   [ "${flavor}" != "citycloud" ] && \
+   [ "${flavor}" != "openstack" ] && \
+   [ "${flavor}" != "aws" ]; then
     log_error "ERROR: Unsupported flavor: ${flavor}"
     exit 1
 fi
@@ -79,6 +83,10 @@ elif [[ "${flavor}" == "gcp" ]]; then
   cp -r "${config_defaults_path}/gcp/group_vars" "${config_path}/"
 elif [[ "${flavor}" == "aws" ]]; then
   cp -r "${config_defaults_path}/aws/group_vars" "${config_path}/"
+elif [[ "${flavor}" == "openstack" ]]; then
+  cp -r "${config_defaults_path}/openstack/group_vars" "${config_path}/"
+elif [[ "${flavor}" == "citycloud" ]]; then
+  cp -r "${config_defaults_path}/openstack/group_vars" "${config_path}/"
 fi
 
 # Copy inventory.ini
--- a/docs/citycloud-snippets.md
+++ b/docs/citycloud-snippets.md
@@ -14,6 +14,13 @@ PUB_SSH_KEY_FILE="${HOME}/.ssh/id_rsa.pub"
 > Save these into a file `my-new-cluster.env`.
 > The next time you want to use these snippets you can just run `source my-new-cluster.env`
 
+Export your cluster path and PGP fingerprint
+
+```bash
+export CK8S_CONFIG_PATH=~/.ck8s/my-new-cluster
+export CK8S_PGP_FP=<Your PGP fingerprint>
+```
+
 Then you'll need to set up some openstack specific variables for the infrastructure
 
 ```bash
@@ -27,12 +34,16 @@ OS_SUBNET_CIDR="10.10.0.0/24" # (Guessing) CIDR of network
 OS_FLOATINGIP_POOL="" # Name of the pool to get floating ips from
 ```
 
+For terraform to be able to talk to the Openstack API server you need to set the appropriate environment variables.
+For more information see [here](https://docs.openstack.org/newton/user-guide/common/cli-set-environment-variables-using-openstack-rc.html).
+
 Set up the clusters into respective folders
 
 ```bash
 pushd kubespray
 for CLUSTER in "${SERVICE_CLUSTER}" "${WORKLOAD_CLUSTERS[@]}"; do
-  mkdir -p "inventory/${CLUSTER}/group_vars"
+  mkdir -p "${CK8S_CONFIG_PATH}/${CLUSTER}-config/group_vars"
+  ln -s "${CK8S_CONFIG_PATH}/${CLUSTER}-config/" "inventory/$CLUSTER"
   # shellcheck disable=SC2016
   sed -e "s@^cluster_name = .*@cluster_name = \"${CLUSTER}\"@" \
       -e "s@^public_key_path = .*@public_key_path = \"${PUB_SSH_KEY_FILE}\"@" \
@@ -80,18 +91,11 @@ If all went well the infrastructure should now be up and running
 
 To set up kubernetes with compliantkubernetes-kubespray you can follow these steps.
 
-Export your cluster path and PGP fingerprint
-
-```bash
-export CK8S_CONFIG_PATH=~/.ck8s/my-new-cluster
-export CK8S_PGP_FP=<Your PGP fingerprint>
-```
-
 Initialize the configuration with.
 
 ```bash
 for CLUSTER in "${SERVICE_CLUSTER}" "${WORKLOAD_CLUSTERS[@]}"; do
-  ./bin/ck8s-kubespray init "${CLUSTER}" citycloud ~/.ssh/id_rsa
+  ./bin/ck8s-kubespray init "${CLUSTER}" openstack ~/.ssh/id_rsa
   ln -s "$(pwd)/kubespray/inventory/${CLUSTER}/tfstate-${CLUSTER}.tfstate" "${CK8S_CONFIG_PATH}/${CLUSTER}-config/" || true
   cp "kubespray/contrib/terraform/openstack/hosts" "${CK8S_CONFIG_PATH}/${CLUSTER}-config/inventory.ini"
   chmod +x "${CK8S_CONFIG_PATH}/${CLUSTER}-config/inventory.ini"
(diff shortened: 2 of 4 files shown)


cristiklein:
Can we add something similar in compliantkubernetes.io?

Ideally, it should only require copying the MD in the right place, updating
`mkdocs.yaml` and checking with `mkdocs serve`.

Xartos:
> Can we add something similar in compliantkubernetes.io? Ideally, it should only require copying the MD in the right place, updating `mkdocs.yaml` and checking with `mkdocs serve`.

Yes, this was my end goal. But I think we might want to actually take some more time to polish this before we add it there. So I thought I just add this here for now so people that need to set it up have something to copy-paste. And then we can create an issue and polish it for compliantkubernetes.io

docs/citycloud-snippets.md
```diff
@@ -0,0 +1,130 @@
...
+OS_SUBNET_CIDR="10.10.0.0/24" # (Guessing) CIDR of network
+OS_FLOATINGIP_POOL="" # Name of the pool to get floating ips from
+```
+
+Set up the clusters into respective folders
+
+```bash
+pushd kubespray
+for CLUSTER in "${SERVICE_CLUSTER}" "${WORKLOAD_CLUSTERS[@]}"; do
+  mkdir -p "inventory/${CLUSTER}/group_vars"
```
lentzi90, line 35: I know this is where the openstack terraform code/docs thinks it should be, but I think it would be better if we could stick to the `$CK8S_CONFIG_PATH`.  My suggestion is to link `contrib/terraform/openstack` into `$CK8S_CONFIG_PATH` or  `$CK8S_CONFIG_PATH/$CLUSTER-config`. This way we keep the terraform state and `cluster.tfvars` under the config path so it is easier to keep track of it.

bin/init.bash
```diff
@@ -30,7 +30,11 @@ if [[ ${CK8S_CLOUD_PROVIDER} != "" ]]; then
 fi
 
 # Validate the flavor
-if [ "${flavor}" != "default" ] && [ "${flavor}" != "gcp" ] && [ "${flavor}" != "aws" ]; then
+if [ "${flavor}" != "default" ] && \
+   [ "${flavor}" != "gcp" ] && \
+   [ "${flavor}" != "citycloud" ] && \
```
lentzi90, line 35: Not sure about this, but let me know what you think. What about skipping citycloud here and just treating it as generic openstack? If we want something specifically for citycloud we can add it later.

lentzi90 (commented):
Really nice job! It is interesting to see the different ways to do this. :smile: 

Added some comments below. Let me know what you think!

docs/citycloud-snippets.md
```diff
@@ -0,0 +1,130 @@
...
+
+Now you're ready to deploy the infrastructure
+
+```bash
+for CLUSTER in "${SERVICE_CLUSTER}" "${WORKLOAD_CLUSTERS[@]}"; do
+  pushd "kubespray/inventory/${CLUSTER}"
+  terraform init ../../contrib/terraform/openstack
+  terraform apply \
+    -var-file cluster.tfvars \
+    -auto-approve \
```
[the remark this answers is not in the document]
Xartos: Yea, I mainly followed the [AWS](https://compliantkubernetes.io/operator-manual/aws/) / [Exoscale](https://github.com/elastisys/compliantkubernetes/pull/41) guide for this code.
Maybe @cristiklein has some thought of this particular decision?

docs/citycloud-snippets.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+> Tip:
+> Save these into a file `my-new-cluster.env`.
+> The next time you want to use these snippets you can just run `source my-new-cluster.env`
+
+Then you'll need to set up some openstack specific variables for the infrastructure
+
+```bash
+OS_IMAGE_NAME="Ubuntu 20.04 Focal Fossa 20200423" # Image name to base VMs on
```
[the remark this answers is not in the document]
Xartos: Yes, I should add a short mentioning of how to authenticate. Maybe link to some documentation.

The reason is mainly because I wanted to, once again,  align with the aws / exoscale guide

bin/init.bash
Xartos: Yes I agree. I mostly wanted to get this here to show my thinking. But since it seems to be the same for safespring (for what we know now) I think we can change it to openstack

docs/citycloud-snippets.md
cristiklein: I prefer this approach. It is closer aligned with kubespray. Terraform state should eventually be handled with `-backend-config`.

docs/citycloud-snippets.md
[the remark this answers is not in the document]
cristiklein: Auto-approve was to simplify copy-paste-ablility. I'm find with eliminating `-auto-approve` as long as the code can be copy-pasted. Last time I checked, this required escaping newline, to make bash "wait" until the very end before starting to ask approve-related questions.

docs/citycloud-snippets.md
```diff
@@ -0,0 +1,130 @@
...
+```bash
+export CK8S_CONFIG_PATH=~/.ck8s/my-new-cluster
+export CK8S_PGP_FP=<Your PGP fingerprint>
+```
+
+Initialize the configuration with.
+
+```bash
+for CLUSTER in "${SERVICE_CLUSTER}" "${WORKLOAD_CLUSTERS[@]}"; do
+  ./bin/ck8s-kubespray init "${CLUSTER}" citycloud ~/.ssh/id_rsa
```
cristiklein, line 94: This is a thorn in my eye. Looking forward to getting rid of it. :smile:

docs/citycloud-snippets.md
```diff
@@ -0,0 +1,130 @@
...
+  ln -s "$(pwd)/kubespray/inventory/${CLUSTER}/tfstate-${CLUSTER}.tfstate" "${CK8S_CONFIG_PATH}/${CLUSTER}-config/" || true
+  cp "kubespray/contrib/terraform/openstack/hosts" "${CK8S_CONFIG_PATH}/${CLUSTER}-config/inventory.ini"
+  chmod +x "${CK8S_CONFIG_PATH}/${CLUSTER}-config/inventory.ini"
```
cristiklein, line 97: I know what you mean here, but it feels weird to make `inventory.ini` executable. Making `hosts` executable feels more expected. Anyway, no need to fix this documentation, but please add an issue to `compliantkubernetes-kubespray` to handle both "static" inventory.ini and "dynamic" hosts.

docs/citycloud-snippets.md
```diff
@@ -0,0 +1,130 @@
...
+done
+```
+
+Done!
+
+## Teardown
+
+Later when you want to destroy the infrastructure.
+
+Make sure all cloud resources are destroyed (persistent volumes, load balancers).
```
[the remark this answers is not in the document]
cristiklein: Is that really that dramatic? At least in AWS, the resources created by Kubernetes are clearly tagged and can easily be removed manually. Maybe we should find the snipped that does so?

cristiklein (approved):

docs/citycloud-snippets.md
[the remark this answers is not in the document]
lentzi90: I see. It is probably fine to keep the auto approve, but I would suggest a note about it to make it more obvious to people. Otherwise I think it may surprise people that are used to always confirming terraform apply.

docs/citycloud-snippets.md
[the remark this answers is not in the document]
lentzi90: I don't think it is that easy to find them in other clouds but I could be wrong. But more importantly, this could cause a surprise bill for people who aren't aware of it. If you know terraform but not kubernetes, it wouldn't be obvious at all that there can be resources left in the cloud after you do `terraform destroy`.

Relevant issue: https://github.com/elastisys/compliantkubernetes-apps/issues/213

docs/citycloud-snippets.md
lentzi90: Ok the state can be handled with a backend config, but what about `cluster.tfvars`? Surely we should keep that in the config folder?

docs/citycloud-snippets.md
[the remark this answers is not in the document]
cristiklein: I see two aspects you bring here:

1. Regarding the Operator Manual, it must start from the assumption that people have CKA or equivalent under the bell. That curriculum does a way better job at explaining the ins-and-outs of Kubernetes than we will ever do. We should focus our documentation on what Compliant Kubernetes *adds* on top of a vanilla Kubernetes cluster, i.e., on what differentiates us, and not redo a half-finished, incomplete, out-of-date version of CKA. :smile: 
2. If the interaction between Kubernetes and underlying cloud resources is not well known or well understood in the community, this would be an excellent blog post.

docs/citycloud-snippets.md
cristiklein: Kubeconfig encourages keeping `cluster.tfvars` in the `inventory/$CLUSTER` folder.

lentzi90: Yes but I don't think it makes sense to have it outside of `$CK8S_CONFIG_PATH`. If we want it in the inventory to stick to the kubespray examples, then I think we should change the config path like this:

```bash
export CK8S_CONFIG_PATH="inventory/${CLUSTER}"
```

cristiklein: I see your point. Let's go with your suggestion.

docs/citycloud-snippets.md
```diff
@@ -0,0 +1,130 @@
+# Citycloud snippets
+
+To set up a cluster on citycloud these snippets can be used
+
+Start by setting up some environments for this setup
+
+```bash
+SERVICE_CLUSTER="my-new-sc-cluster"
+WORKLOAD_CLUSTERS=( "my-new-wc-cluster" )
+PUB_SSH_KEY_FILE="${HOME}/.ssh/id_rsa.pub"
```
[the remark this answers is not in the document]
Xartos: Not at all. But shellcheck complained about it so I changed it

docs/citycloud-snippets.md
Xartos: This is a requirement for kubespray. Since it couldn't execute this script without having permission to execute it

bin/init.bash
Xartos: You certainly have a point there

bin/init.bash
```diff
@@ -79,6 +83,10 @@ elif [[ "${flavor}" == "gcp" ]]; then
   cp -r "${config_defaults_path}/gcp/group_vars" "${config_path}/"
 elif [[ "${flavor}" == "aws" ]]; then
   cp -r "${config_defaults_path}/aws/group_vars" "${config_path}/"
+elif [[ "${flavor}" == "openstack" ]]; then
+  cp -r "${config_defaults_path}/openstack/group_vars" "${config_path}/"
+elif [[ "${flavor}" == "citycloud" ]]; then
+  cp -r "${config_defaults_path}/openstack/group_vars" "${config_path}/"
```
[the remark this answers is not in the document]
Xartos: Hmm, would that work. Since when we copy these files the structure changes so the links would need to change. 

If I understand you correctly the structure would look something like this:
```
openstack/group_vars/some_openstack_file.txt
citycloud/group_vars/some_citycloud_file.txt -> ../../openstack/group_vars/some_openstack_file.txt
```
And after you copied the citycloud config it would look like this:
```
group_vars/some_citycloud_file.txt -> ../../openstack/group_vars/some_openstack_file.txt
```
Or am I missing something?

docs/citycloud-snippets.md
cristiklein: I get that the script needs to be executable for kubespray. However, I would prefer that executable to be called `hosts` (without extension) and not `something.ini`. The `ini` part feels misleading.

bin/init.bash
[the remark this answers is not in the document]
cristiklein: Nice catch @Xartos ! I think you would need to use `cp --dereference`. I think this is already done some in other parts of kubespray.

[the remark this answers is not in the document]
Xartos: Right, yea then I agree. That is a quite neat solution

docs/citycloud-snippets.md
Xartos: Right, yes that is a bit misleading. And yes, creating a new issue for this is probably the best for now.

bin/init.bash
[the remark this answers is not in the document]
Xartos: Actually, I think that this change might go a bit too far for this issue. I'll create a new issue for this instead.

lentzi90 (approved):
LGTM

PR #47, opened 2021-02-17
bug: kubeconfig now copied to .state dir after install

### LINKED ISSUE #37: [1] kube_config is not copied to config dir
**Describe the bug**
After kubespray has completed creating a cluster, admin.conf should be copied to .state/kube_config_[cluster].yaml However, currently this gives an error due to the .state directory not existing.

**To Reproduce**
This happens during the ./bin/ck8s-kubespray apply $CLUSTER stage

**Expected behavior**
the kube_config files should be available in the .state dir

pettersv:
Fixes #37

After kubespray has completed creating a cluster, admin.conf should be copied to .state/kube_config_[cluster].yaml However, currently this gives an error due to the .state directory not existing.

This PR creates that dir if it doesnt exist.

### DIFF
--- a/bin/apply.bash
+++ b/bin/apply.bash
@@ -29,6 +29,7 @@ popd
 log_info "Kubespray done"
 
 if [ -f "${config_path}/artifacts/admin.conf" ]; then
+    mkdir -p "${state_path}"
     mv "${config_path}/artifacts/admin.conf" "${secrets[kube_config]}"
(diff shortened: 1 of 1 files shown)


Ajarmar (approved):
Looks good to me (although I haven't run into this issue myself - does it only occur under certain circumstances?)

viktor-f (approved):

PR #62, opened 2021-03-09
bin: add apply-ssh command to change SSH keys

Ajarmar:
**What this PR does / why we need it**: Adds a command to the `ck8s-kubespray` wrapper that enables changing authorized SSH keys for a cluster.

**Which issue this PR fixes**: fixes an issue in an internal repository

**Checklist:**

  - [X] is completely transparent, will not impact the workload in any way.
  - [ ] will break the cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Content:
 
     ```bash
     export CK8S_CONFIG_PATH=~/.ck8s/my-environment
-    ./bin/ck8s-kubespray init <prefix> <flavor> <path to ssh key> [<SOPS fingerprint>]
+    ./bin/ck8s-kubespray init <prefix> <flavor> [<SOPS fingerprint>]
     ```
 
     Arguments:
@@ -47,6 +47,21 @@ Content:
 1. Done.
    You should now have a working kubernetes cluster.
    You should also have an encrypted kubeconfig at `<CK8S_CONFIG_PATH>/.state/kube_config_<prefix>.yaml` that you can use to access the cluster.
+
+## Changing authorized SSH keys for a cluster
+
+Authorized SSH keys can be changed for a cluster using:
+
+```bash
+./bin/ck8s-kubespray apply-ssh <prefix> [<options>]
+```
+
+It will set the public SSH key(s) found in`<CK8S_CONFIG_PATH>/<prefix>-config/group_vars/all/ck8s-ssh-keys.yaml` as authorized keys in your cluster (just add the keys you want to be authorized as elements in `ck8s_ssh_pub_keys_list`).
+Note that the authorized SSH keys for the cluster will be set to these keys _exclusively_, removing any keys that may already be authorized, so make sure the list includes **every SSH key** that should be authorized.
+
+When running this command, the SSH keys are applied to each node in the cluster sequentially, in reverse inventory order (first the workers and then the masters).
+A connection test is performed after each node which has to succeed in order for the playbook to continue.
+If the connection test fails, you may have lost your SSH access to the node; to recover from this, you can set up an SSH connection before running the command and keep it active so that you can change the authorized keys manually.
 
 ## Running other kubespray playbooks
 
--- a/bin/apply-ssh.bash
+++ b/bin/apply-ssh.bash
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# This script will run an ansible playbook.
+# It's not to be executed on its own but rather via `ck8s-kubespray apply-ssh <prefix>`.
+
+set -eu -o pipefail
+
+here="$(dirname "$(readlink -f "$0")")"
+# shellcheck source=common.bash
+source "${here}/common.bash"
+
+log_info "Running playbook authorized_key"
+pushd "${here}/../playbooks"
+
+ansible-playbook -i "${config[inventory_file]}" authorized_key.yml "$@"
+
+popd
+
+log_info "Playbook ran successfully!"
--- a/playbooks/authorized_key.yml
+++ b/playbooks/authorized_key.yml
@@ -1,0 +1,22 @@
+- hosts: all
+  serial: 1
+  order: reverse_inventory
+  tasks:
+    - name: Check that the SSH key list file exists
+      assert:
+        that: ck8s_ssh_pub_keys_list is defined
+        fail_msg: "ck8s_ssh_pub_keys_list is undefined (the group_vars file is likely missing). Did you run ck8s-kubespray init?"
+    - name: Check that SSH key list is valid
+      assert:
+        that: ck8s_ssh_pub_keys_list and None not in ck8s_ssh_pub_keys_list
+        fail_msg: "Public SSH key list invalid - add SSH keys to ck8s_ssh_pub_keys_list"
+    - name: Add authorized keys from config folder
+      ansible.posix.authorized_key:
+        user: "{{ ansible_user }}"
+        state: present
+        exclusive: true
+        key: "{{ ck8s_ssh_pub_keys_list | join('\n') }}"
+    - name: Test connection
+      wait_for_connection:
+        delay: 1
+        timeout: 30
--- a/bin/ck8s-kubespray
+++ b/bin/ck8s-kubespray
@@ -15,6 +15,8 @@ usage() {
     echo "      args: <prefix> [<options>]" 1>&2
     echo "  run-playbook                                runs any ansible playbook in kubespray" 1>&2
     echo "      args: <prefix> <playbook> [<options>]" 1>&2
+    echo "  apply-ssh                                   applies SSH keys from a file to a cluster" 1>&2
+    echo "      args: <prefix> [<options>]" 1>&2
     exit 1
 }
 
@@ -48,6 +50,13 @@ case "${1}" in
         fi
         shift 2
         "${here}/run-playbook.bash" "${@}"
+        ;;
+    apply-ssh)
+        if [ $# -lt 2 ]; then
+            usage
+        fi
+        shift 2
+        "${here}/apply-ssh.bash" "${@}"
         ;;
     *) usage ;;
 esac
--- a/config/common/group_vars/all/ck8s-ssh-keys.yaml
+++ b/config/common/group_vars/all/ck8s-ssh-keys.yaml
@@ -1,0 +1,2 @@
+ck8s_ssh_pub_keys_list:
+  -


README.md
```diff
@@ -48,6 +48,19 @@ Content:
    You should now have a working kubernetes cluster.
    You should also have an encrypted kubeconfig at `<CK8S_CONFIG_PATH>/.state/kube_config_<prefix>.yaml` that you can use to access the cluster.
 
+## Changing authorized SSH keys for a cluster
+
+Authorized SSH keys can be changed for a cluster using `ck8s-kubespray apply-ssh`. The prerequisites for doing this are:
+
+- A file containing the public SSH keys that should be authorized to access the cluster, separated by line breaks. Note that the authorized SSH keys for the cluster will be set to these keys _exclusively_, removing any keys that may already be authorized, so make sure the file includes **every SSH key** that should be authorized.
+- Set the environment variable `SSH_KEY_LOCATION` to the absolute path of the file.
```
cristiklein, line 56: I'm not sure how to deal with this in a multi-VM setting, but for single VMs, we used to keep a "backup" SSH connection, just in case we cut ourselves off.

Is this still recommendable today? My thinking is that in worst case, you can always fix an SSH key misconfiguration by re-creating the VMs and putting the right keys in `cloud-init`.

cristiklein (approved):

README.md
Xartos: I'm not sure how this is done in the wild, but could it be an idea to have the script do it one-by-one and test the ssh connection between every change? If that's possible

cristiklein: Certainly possible, but is it worth the effort? :smile:

playbooks/authorized_key.yml
```diff
@@ -0,0 +1,10 @@
+- hosts: all
+  become: yes
+  become_user: ubuntu
+  tasks:
+    - name: Add authorized keys from $SSH_KEY_LOCATION
+      ansible.posix.authorized_key:
+        user: ubuntu
```
lentzi90, line 7: This would also go away or be changed to `user: "{{ ansible_user }}"`

bin/apply-ssh.bash
```diff
@@ -0,0 +1,23 @@
...
+: "${SSH_KEY_LOCATION:?Missing SSH_KEY_LOCATION}"
+
+here="$(dirname "$(readlink -f "$0")")"
+# shellcheck source=common.bash
+source "${here}/common.bash"
+
+log_info "Running playbook authorized_key"
+pushd "${here}/../playbooks"
+
+ansible-playbook -i "${config[inventory_file]}" authorized_key.yml -b --extra-vars "ssh_key_location=$SSH_KEY_LOCATION" "${*}"
```
lentzi90, line 19: Skip the `-b`, no need to be root for this, right?
```suggestion
ansible-playbook -i "${config[inventory_file]}" authorized_key.yml --extra-vars "ssh_key_location=$SSH_KEY_LOCATION" "${*}"
```

bin/apply-ssh.bash
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# This script will run an ansible playbook available in kubespray.
```
lentzi90, line 3:
```suggestion
# This script will run an ansible playbook.
```

lentzi90 (changes requested):
Great start!
I have a few suggestions to make it more flexible.

bin/apply-ssh.bash
```diff
@@ -0,0 +1,23 @@
...
+# It will also install some python dependencies for kubespray in a virtual environment
```
Xartos, line 4:
```suggestion
```

bin/apply-ssh.bash
```diff
@@ -0,0 +1,23 @@
...
+# It's not to be executed on its own but rather via `ck8s-kubespray apply-ssh <prefix>`.
+
+set -eu -o pipefail
+shopt -s globstar nullglob dotglob
```
Xartos, line 8: Is this needed?

bin/apply-ssh.bash
```diff
@@ -0,0 +1,23 @@
...
+
+: "${SSH_KEY_LOCATION:?Missing SSH_KEY_LOCATION}"
```
Xartos, line 10:
```suggestion
: "${SSH_KEY_LOCATION:?Missing SSH_KEY_LOCATION}"
if [ ! -f "${SSH_KEY_LOCATION}" ]; then
```
Or maybe ansible will fail if it's no file?

Xartos (commented):
Great stuff! just some comments,

README.md
lentzi90: This should be possible with [serial](https://docs.ansible.com/ansible/latest/user_guide/playbooks_strategies.html#setting-the-batch-size-with-serial) or [throttle](https://docs.ansible.com/ansible/latest/user_guide/playbooks_strategies.html#restricting-execution-with-throttle) together with [step](https://docs.ansible.com/ansible/2.9/user_guide/playbooks_startnstep.html#step) or possibly a [confirmation step](https://stackoverflow.com/questions/34290525/pause-ansible-playbook-for-user-confirmation-whether-to-run-rest-tasks) I think.
Alternatively the [debug strategy](https://docs.ansible.com/ansible/latest/user_guide/playbooks_debugger.html#enabling-the-debugger-as-a-strategy) maybe?

bin/apply-ssh.bash
Ajarmar: Yeah, ansible fails if it's not a file. It seems cleaner to put a warning at this point though, so I'll add it

bin/apply-ssh.bash
cristiklein: I thought about this and I don't like it. This repo is a 95% Ansible solution, so we should avoid technology sprawl. Let's just add an Ansible variable `ssh_public_keys` which is stored the same way all the other variables are stored.

Ajarmar: We don't really have an established way of handling variables for ansible playbooks in this repo, since all playbooks we use are in the upstream kubespray repo. I think the easiest way of avoiding this extra environment variable would be to make this into a proper ansible role instead of just having the playbook, or do you have a different suggestion?

cristiklein: We currently have `./bin/ck8s apply sc`, which does:

```
ansible-playbook -i ${config[inventory_file]} ${playbook} -b ${*}
```

If playbook was `.../authorized_key.yml`, with `ssh_public_keys` taken from `$CK8S_CONFIG_PATH/$CLUSTER`, then I would be very happy. I don't feel we need a full-blown role, just for importing `ansible.posix.authorized_key`.

lentzi90: I guess what @Ajarmar is trying to say is that we so far have used *only kubespray variables* for ansible. We have not defined any new variables or even used ansible for anything other than kubespray.

If we add our own variables (which I think makes sense) we should either use a separate inventory (not so nice) or make sure we separate the variables clearly from what kubespray uses so we don't risk conflicts. I think the best way to achieve this is to explicitly read the variables from a file outside of `group_vars` like [described here](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#defining-variables-in-included-files-and-roles). SO in the playbook we would have something like this:

```yaml
- hosts: all
  vars_files:
    - /path/to/ck8s/specific/external_vars.yml
```

cristiklein: I see your point. I understand the need to avoid name clashes with upstream Kubespray. However, having a separate file feels overkill. I would simply use something like the `ck8s_` prefix and agree that those are "our" variables (so `ck8s_ssh_public_keys` ?). We are likely to add more variables for security hardening CK8s clusters in the future.

Ajarmar: I feel like putting these kinds of variables in their own file makes more sense. [The only common config files we have for every cloud provider](https://github.com/elastisys/compliantkubernetes-kubespray/tree/main/config/common/group_vars) are `etcd` and `k8s-cluster`. Putting a variable that only affects something locally on your computer (i.e. where you look for the SSH keys you want to apply) among the Kubernetes configuration variables seems a bit weird. If we add our own variables that actually affect the cluster directly in some way then sure, but that's not really the case here.

cristiklein: I would definitely not *point* to a file with all the SSH public keys, rather *list all public SSH keys* in the configuration file. Pointing to a configuration file that is outside git is a recipe for disaster. See previous discussion that we had about SSH public keys in the `cloud-init` context: https://github.com/kubernetes-sigs/kubespray/pull/7242

OlleLarsson: Yeah, I think having an ansible variables file in each cluster config repo sound good.
Somehting similar to
```ShellSession
# ${CK8S_CONFIG_PATH}/ssh_keys.yml
ssh_pub_keys_list:
  - ssh-rsa AAAAB3Nza1
  - ssh-rsa AAAAB3Nza2
  - ...
```

README.md
```diff
@@ -25,6 +25,8 @@ Content:
       Supported options are `default`, `gcp`, `aws`, and `openstack`.
     * `SOPS fingerprint` is the gpg fingerprint that will be used for SOPS encryption.
       You need to set this or the environment variable `CK8S_PGP_FP` the first time SOPS is used in your specified config path.
+    * Additionally, the init script will copy your public SSH key from `~/.ssh/id_rsa.pub` to the config path.
+      If you want a different SSH key file copied, set the environment variable `SSH_KEY_LOCATION` to the absolute path of the file. The file can contain several SSH keys, one per line.
```
Xartos, line 29: I'm not a fan of mixing envs and positional arguments. Is there any particular reason for why we settle on env var for this? I also think that we shouldn't make this optional and default to `~/.ssh/id_rsa.pub`, better to have the user explicitly choose. Then you're less prone to accidentally only include your own ssh key.

Thoughts on this?

README.md
```diff
@@ -48,6 +50,18 @@ Content:
...
 
+## Changing authorized SSH keys for a cluster
+
+Authorized SSH keys can be changed for a cluster using:
+
+```bash
+./bin/ck8s-kubespray apply-ssh <prefix>
+```
+
+It will set the public SSH key(s) found in`<CK8S_CONFIG_PATH>/<prefix>-config/ssh-keys.pub` as authorized keys in your cluster (to apply multiple keys, put one key per line in the file). Note that the authorized SSH keys for the cluster will be set to these keys _exclusively_, removing any keys that may already be authorized, so make sure the file includes **every SSH key** that should be authorized.
```
Xartos, line 61:
```suggestion
It will set the public SSH key(s) found in`<CK8S_CONFIG_PATH>/<prefix>-config/ssh-keys.pub` as authorized keys in your cluster (to apply multiple keys, put one key per line in the file).
Note that the authorized SSH keys for the cluster will be set to these keys _exclusively_, removing any keys that may already be authorized, so make sure the file includes **every SSH key** that should be authorized.
```

README.md
```diff
@@ -48,6 +50,18 @@ Content:
...
+
+When running this command, the SSH keys are applied to each node in the cluster sequentially, first the master and then the workers. A connection test is performed after each node which has to succeed in order for the playbook to continue. If the connection test fails, you may have lost your SSH access to the node; to recover from this, you can set up an SSH connection before running the command and keep it active so that you can change the authorized keys manually.
```
Xartos, line 63:
```suggestion
When running this command, the SSH keys are applied to each node in the cluster sequentially, first the master and then the workers.
A connection test is performed after each node which has to succeed in order for the playbook to continue.
If the connection test fails, you may have lost your SSH access to the node; to recover from this, you can set up an SSH connection before running the command and keep it active so that you can change the authorized keys manually.
```

bin/apply-ssh.bash
```diff
@@ -0,0 +1,24 @@
...
+# This script will run an ansible playbook.
+# It's not to be executed on its own but rather via `ck8s-kubespray apply-ssh <prefix>`.
+
+set -eu -o pipefail
+
+here="$(dirname "$(readlink -f "$0")")"
+# shellcheck source=common.bash
+source "${here}/common.bash"
+
+if [ ! -f "${config_path}"/ssh-keys.pub ]; then
```
Xartos, line 12: Just to make sure it isn't an empty file.
```suggestion
if [ ! -f "${config_path}"/ssh-keys.pub ] || [ ! -s "${config_path}"/ssh-keys.pub ]; then
```

Xartos (commented):
Very nice, much wow! :dog: Although I have some comments

README.md
Ajarmar: The init script doesn't really allow for several optional arguments at the moment (it just checks for the SOPS fingerprint in a specific position) and I wasn't sure if this was the moment to change it into something that works better (i.e. specifying flags like `-sops=... -ssh=...`). But as you say, maybe we want to make it mandatory anyway, the only thing I don't like about that is the case where the user just wants to use their `~/.ssh/id_rsa.pub` and they still have to specify the path to it - feels a bit redundant.

README.md
lentzi90, line 61: I think it would make more sense to call the file `authorized_hosts`. This is how it is "usually done" with ssh and it goes well together with the rest of the text. :slightly_smiling_face: 
```suggestion
It will set the public SSH key(s) found in`<CK8S_CONFIG_PATH>/<prefix>-config/authorized_hosts` as authorized keys in your cluster (to apply multiple keys, put one key per line in the file). Note that the authorized SSH keys for the cluster will be set to these keys _exclusively_, removing any keys that may already be authorized, so make sure the file includes **every SSH key** that should be authorized.
```

bin/apply-ssh.bash
lentzi90, line 12:
```suggestion
if [ ! -f "${config_path}"/authorized_hosts ]; then
```

bin/apply-ssh.bash
```diff
@@ -0,0 +1,24 @@
...
+  log_error "ERROR: ${config_path}/ssh-keys.pub not found. Did you run ck8s-kubespray init?"
```
lentzi90, line 13:
```suggestion
  log_error "ERROR: ${config_path}/authorized_hosts not found. Did you run ck8s-kubespray init?"
```

bin/apply-ssh.bash
```diff
@@ -0,0 +1,24 @@
...
+  exit 1
+fi
+
+log_info "Running playbook authorized_key"
+pushd "${here}/../playbooks"
+
+ansible-playbook -i "${config[inventory_file]}" authorized_key.yml --extra-vars "ssh_key_location=${config_path}/ssh-keys.pub" "$@"
```
lentzi90, line 20:
```suggestion
ansible-playbook -i "${config[inventory_file]}" authorized_key.yml --extra-vars "ssh_key_location=${config_path}/authorized_hosts" "$@"
```

bin/init.bash
```diff
@@ -50,6 +56,15 @@ generate_sops_config() {
     sops_config_write_fingerprints "${fingerprint}"
 }
 
+copy_public_ssh_key() {
+  if [ "$SSH_KEY_LOCATION" == "" ] || [ ! -f "$SSH_KEY_LOCATION" ]; then
+    log_info "SSH_KEY_LOCATION unset, copying ~/.ssh/id_rsa.pub"
+    cp ~/.ssh/id_rsa.pub "${config_path}"/ssh-keys.pub
+  else
+    cp "$SSH_KEY_LOCATION" "${config_path}"/ssh-keys.pub
```
lentzi90, line 64:
```suggestion
    cp ~/.ssh/id_rsa.pub "${config_path}"/authorized_hosts
  else
    cp "$SSH_KEY_LOCATION" "${config_path}"/authorized_hosts
```

lentzi90 (approved):
Nice!
Just a minor thing: I suggest changing the name of the file from `id_rsa.pub` to `authorized_hosts`.

Ajarmar:
@cristiklein I made changes according to our discussion about making it more "standard ansible", please check that you think it looks fine.

@lentzi90 I made some changes to the files that are used, do you think it still would be good to make this naming change? Should the ansible variable (currently `ck8s_ssh_pub_keys_list`) also be renamed in that case?

cristiklein (approved):
I love this! Idiot-proof and perfectly in line with what a person skilled in Ansible would expect.

README.md
```diff
@@ -48,6 +48,21 @@ Content:
...
+Authorized SSH keys can be changed for a cluster using:
+
+```bash
+./bin/ck8s-kubespray apply-ssh <prefix> [<options>]
+```
+
+It will set the public SSH key(s) found in`<CK8S_CONFIG_PATH>/<prefix>-config/group_vars/all/ck8s-ssh-keys.yaml` as authorized keys in your cluster (just add the keys you want to be authorized as elements in `ck8s_ssh_pub_keys_list`).
+Note that the authorized SSH keys for the cluster will be set to these keys _exclusively_, removing any keys that may already be authorized, so make sure the list includes **every SSH key** that should be authorized.
+
+When running this command, the SSH keys are applied to each node in the cluster sequentially, first the master and then the workers.
```
lentzi90, line 62: I think it would be pretty nice to run first on workers and then control-plane nodes. Otherwise, if something goes wrong you will lose a control-plane node. Worst case: you have a single control-plane node so you loose the whole cluster. :scream:

lentzi90 (approved):
Great!
I think the naming works well like this, no need to change it. If you still have the time I think it would be nice to control the [order](https://docs.ansible.com/ansible/latest/user_guide/playbooks_strategies.html#ordering-execution-based-on-inventory) so that workers are done first.

README.md
Ajarmar: Yeah, this seems smart. I changed it.

lentzi90 (approved):
LGTM!

PR #66, opened 2021-03-18
Replace ~ with $HOME so absolute path expansion works

Xartos:
Since `~` isn't expanded by `readlink` when it's in a variable I would suggest this small fix to make us able to use `~` to reference `${HOME}`

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: 
N/A

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/bin/common.bash
+++ b/bin/common.bash
@@ -7,9 +7,17 @@
 : "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
 : "${prefix:?Missing prefix}"
 
+# Check for this mistake https://github.com/koalaman/shellcheck/wiki/SC2088
+# shellcheck disable=SC2088
+if [[ "${CK8S_CONFIG_PATH:0:2}" == "~/" ]]; then
+    echo "Warning: CK8S_CONFIG_PATH starts with unexpanded ~/" 1>&2
+    echo "This will create a new folder in cwd called ~ instead of referencing ${HOME}" 1>&2
+    echo "please use \${HOME} instead if that's what you want" 1>&2
+fi
+
 # Create CK8S_CONFIG_PATH if it does not exist and make it absolute
-mkdir -p "${CK8S_CONFIG_PATH}"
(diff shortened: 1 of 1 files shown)


cristiklein:
I don't get it:

```
$ export CK8S_CONFIG_PATH=~/.aws
$ echo $CK8S_CONFIG_PATH
/home/cklein/.aws
```

Isn't `~` expansion supposed to happen when assigning `CK8S_CONFIG_PATH` in the shell, i.e., before calling any of our scripts?

Xartos:
> I don't get it:
> 
> ```
> $ export CK8S_CONFIG_PATH=~/.aws
> $ echo $CK8S_CONFIG_PATH
> /home/cklein/.aws
> ```
> 
> Isn't `~` expansion supposed to happen when assigning `CK8S_CONFIG_PATH` in the shell, i.e., before calling any of our scripts?

TIL: You are correct, as long as you don't put it in quotes

```console
$ CK8S_CONFIG_PATH="~/.ck8s/test"
$ echo ${CK8S_CONFIG_PATH}
~/.ck8s/test

$ CK8S_CONFIG_PATH=~/.ck8s/test
$ echo ${CK8S_CONFIG_PATH}
/home/freddeliv/.ck8s/test
```

So if you put your path in quotes our script breaks which is why I haven't encountered this before. 

But this would catch that cornercase

cristiklein:
Sorry to play devil's advocate, but is this really the right layer to fix a user error? What if I want to name my folder `ck8s~customer1`? What if I want to refer to another user's folder via `~user2/.ck8s/my-first-cluster`?

Can we display an actionable and informative warning to the user instead? "WARN: '~' character detected in CK8S_CONFIG_PATH. If you intended to refer to your home folder, please unquote it to let the shell expand it.".

Xartos:
Yes, you are right, a warning is probably better. This solution made more sense when I thought it was always the case.
But I will test to add a warning and then we can see how that feels :+1:

cristiklein (approved):

OlleLarsson (approved):

PR #68, opened 2021-03-23
add more openstack variables

viktor-f:
**What this PR does / why we need it**:Added some more variables for openstack, primarily related to enabling kubernetes loadbalancer services.

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/config/openstack/group_vars/k8s-cluster/ck8s-k8s-cluster-openstack.yaml
+++ b/config/openstack/group_vars/k8s-cluster/ck8s-k8s-cluster-openstack.yaml
@@ -2,3 +2,26 @@ etcd_kubeadm_enabled: true
 
 cloud_provider: openstack
 calico_mtu: 1480
+
+## The following variables are used to enable loadbalancer services in kubernetes.
+## All of them might not be needed in your case.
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
Not sure how I feel about this. On the one hand it is nice to have relevant comments when you initialize for a specific cloud, but on the other hand I could just look up these variables in the kubespray documentation and set them. I guess these comments will easily become deprecated also if/when we forget to update them and kubespray changes.

OlleLarsson:
Maybe we could note configuration like this is in our documentation in the ops repo?
https://github.com/elastisys/ck8s-ops/blob/main/docs/new-openstack.md

viktor-f:
Hmm, I see your point. Could another option be to just add a note like this (instead of copying the variables like I did) to our group vars (and also to the `new-openstack.md` doc):
 ```
## More openstack variables (e.g. to enable loadbalancer services) can be found here: https://github.com/kubernetes-sigs/kubespray/blob/master/inventory/sample/group_vars/all/openstack.yml
```

lentzi90:
Comment and note in docs sounds good!

PR #69, opened 2021-03-29
Small fixes: dynamic inventory path; don't force become

### LINKED ISSUE #31: [1] Multiple clusters with dynamic inventory
This can be seen as a subtask or replacement of https://github.com/elastisys/ck8s-ops/issues/57.

The [terraform code for openstack](https://github.com/kubernetes-sigs/kubespray/tree/master/contrib/terraform/openstack) among others, make use of a [python script for dynamic inventories](https://github.com/kubernetes-sigs/kubespray/blob/master/contrib/terraform/terraform.py).
This works well for a single cluster, but as soon as you try to set up a second, you will notice that it finds the terraform state of *all* clusters and tries to create one huge cluster instead of two separate.

The script works by searching for `*.tfstate` "below the working directory".
I'm not sure what exactly the working directory is when [using compliantkubernetes-kubespray](https://github.com/elastisys/compliantkubernetes-kubespray/blob/main/bin/apply.bash#L25), but for me at least it finds the state in both `sc-config` and `wc-config`, directly below my `$CK8S_CONFIG_PATH`.

Definition of done: It is possible to use compliantkubernetes-kubespray with multiple clusters using dynamic inventories.

### LINKED ISSUE #61: Don't force "become"
We currently always set the [become](https://github.com/elastisys/compliantkubernetes-kubespray/blob/main/bin/run-playbook.bash#L32) flag `-b` when running playbooks. However, not all playbooks require it and it could be surprising or even unwanted. For example, the [mitogen playbook](https://github.com/kubernetes-sigs/kubespray/blob/a923f4e7c0692229c442b07a531bfb5fc41a23f9/mitogen.yml) just installs an ansible plugin in the kubespray directory and does not need root for anything. By forcing become, we make that plugin owned by root.

I would like to simply remove `-b` from the `run-playbook` command. For `apply` I think it is fine.

We can continue as it is now. I'm not sure about the implications of installing the mitogen plugin as root.

**Additional context**
[Docs for the mitogen playbook](https://github.com/kubernetes-sigs/kubespray/blob/a923f4e7c0692229c442b07a531bfb5fc41a23f9/docs/mitogen.md).

**Definition of done:**
You can use the `run-playbook` command without the "become" flag.

lentzi90:
This fixes a couple of small issues.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #31, fixes #61

I'm not sure if it is a good idea to set `TERRAFORM_STATE_ROOT` in `common.bash` or if it would be better to set it in the `run-playbook.bash` and `apply.bash` scripts instead. What do you think?

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/bin/common.bash
+++ b/bin/common.bash
@@ -31,6 +31,9 @@ config_path="${CK8S_CONFIG_PATH}/${prefix}-config"
 sops_config="${CK8S_CONFIG_PATH}/.sops.yaml"
 state_path="${CK8S_CONFIG_PATH}/.state"
 
+# Set the path to search for dynamic inventories
+export TERRAFORM_STATE_ROOT="${config_path}"
+
 declare -A config
 # shellcheck disable=SC2034
 config["inventory_file"]="${config_path}/inventory.ini"
--- a/bin/apply.bash
+++ b/bin/apply.bash
@@ -22,7 +22,7 @@ source venv/bin/activate
 pip install -r requirements.txt
 
 log_info "Running kubespray"
-ansible-playbook -i "${config[inventory_file]}" cluster.yml -b ${*}
+ansible-playbook -i "${config[inventory_file]}" cluster.yml -b "${@}"
 
 popd
 
--- a/bin/run-playbook.bash
+++ b/bin/run-playbook.bash
@@ -29,7 +29,7 @@ source venv/bin/activate
 pip install -r requirements.txt
 
 log_info "Running kubespray"
-ansible-playbook -i ${config[inventory_file]} ${playbook} -b ${*}
+ansible-playbook -i "${config[inventory_file]}" "${playbook}" "${@}"
 
 popd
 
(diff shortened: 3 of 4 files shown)


Xartos (approved):
LGTM!

bin/common.bash
```diff
@@ -31,6 +31,9 @@ config_path="${CK8S_CONFIG_PATH}/${prefix}-config"
 sops_config="${CK8S_CONFIG_PATH}/.sops.yaml"
 state_path="${CK8S_CONFIG_PATH}/.state"
 
+# Set the path to search for dynamic inventories
+export TERRAFORM_STATE_ROOT="${config_path}"
```
cristiklein, line 35: Does this need documentation update?

cristiklein (approved):

bin/common.bash
lentzi90: I don't think so. Previous behavior was surprising: it found the state of all clusters under `CK8S_CONFIG_PATH`, so essentially trying to build a mega cluster by combining SC and all WCs. If you happened to know about this variable you could avoid it but it isn't really documented.
Now it just does what you would expect.

Edit: Seems like it is actually used in a snippet without explanation. Will remove it from there. :slightly_smiling_face:

PR #73, opened 2021-04-15
config: Update kubernetes to include security patch

Xartos:

Update kubernetes version to include security patch.
Will require update on already running clusters.

It might be an idea to do a release of this repo before this PR is merged

**Checklist:**

  - [x] requires running a migration script.
  - [ ] will break the cluster.

### DIFF
--- a/config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
+++ b/config/common/group_vars/k8s-cluster/ck8s-k8s-cluster.yaml
@@ -3,6 +3,8 @@ kube_oidc_url: "https://dex.BASE-DOMAIN"
 kube_oidc_client_id: "kubelogin"
 kube_oidc_username_claim: "email"
 
+kube_version: v1.19.10
+
 kube_apiserver_enable_admission_plugins:
   - "PodSecurityPolicy"
   - "NamespaceLifecycle"


viktor-f (dismissed):
LGTM! And yes a release before this would be good.

lentzi90 (commented):
I don't think this will work unless you also update kubespray. You need to pick a version that [it knows about](https://github.com/kubernetes-sigs/kubespray/blob/a923f4e7c0692229c442b07a531bfb5fc41a23f9/roles/download/defaults/main.yml).

Xartos:
> I don't think this will work unless you also update kubespray. You need to pick a version that [it knows about](https://github.com/kubernetes-sigs/kubespray/blob/a923f4e7c0692229c442b07a531bfb5fc41a23f9/roles/download/defaults/main.yml).

Right I forgot about that... Well I guess we'd just have to wait (or fix it ourselves)

cristiklein:
Seems to be the case on `master`: https://github.com/kubernetes-sigs/kubespray/blob/master/roles/download/defaults/main.yml#L213

Would you like to cherry-pick this to `release-2.15`?

Xartos:
> Seems to be the case on `master`: https://github.com/kubernetes-sigs/kubespray/blob/master/roles/download/defaults/main.yml#L213
> 
> Would you like to cherry-pick this to `release-2.15`?

Oh, I was searching in PRs and issues but didn't find anything about this. I guess githubs search isn't always optimal.
Yea, I'll do that

lentzi90:
Actually, if we just update the the latest commit on the `release-2.15` we already get newer versions. See https://github.com/kubernetes-sigs/kubespray/blob/release-2.15/roles/download/defaults/main.yml#L195

Xartos:
> we already get newer versions

Yea, but we want `1.19.10` (the version that fixes the security issue)
Or that was my intention with this PR.

Xartos:
We can now use this in from the branch `release-2.15` https://github.com/kubernetes-sigs/kubespray/pull/7517

viktor-f:
With the upgrade to kubespray 2.16 we are now using k8s 1.20.7, so this is no longer relevant.

PR #76, opened 2021-04-23
Allow skipping venv and requirements installation 

simonklb:

Installing the Kubespray requirements in a virtual environment is great but in certain circumstances you might want to rely on everything being installed in your current environment / system. A good example would be a pre-built Docker image where the current Kubespray version we target is already installed.

I'm not set on the `CK8S_KUBESPRAY_NO_VENV` env name. Feel free to suggest a different, more fitting one.

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/bin/apply.bash
+++ b/bin/apply.bash
@@ -15,11 +15,13 @@ log_info "Creating kubernetes cluster using kubespray"
 # shellcheck disable=SC2154
 pushd "${kubespray_path}"
 
-log_info "Installing requirements for kubespray"
-python3 -m venv venv
-# shellcheck source=../kubespray/venv/bin/activate
-source venv/bin/activate
-pip install -r requirements.txt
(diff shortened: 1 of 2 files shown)


Xartos (commented):
:+1: Could you add it to `run-playbook` as well?

lentzi90 (approved):
LGTM!
Should we mention this in the readme? Or maybe wait until it has been tested a bit more?

simonklb:
>  Could you add it to `run-playbook` as well?

Done!

simonklb:
> LGTM!
> Should we mention this in the readme? Or maybe wait until it has been tested a bit more?

I'm fine with leaving it out for now since it's completely optional. Up to you!

lentzi90:
Then let's merge it like this! :slightly_smiling_face: We can document it if we find it being used a lot later.

Xartos (approved):
LGTM

PR #86, opened 2021-05-31
bin/init: Use cloud provider instead of flavor

simonklb:

This aligns better with compliantkubernetes-apps in two ways:
* ck8s-apps uses `CK8S_CLOUD_PROVIDER`, you will have to set/unset it between runs if you try to use both.
* ck8s-apps uses the term "flavor" to indicate different kinds of the same thing, e.g. cluster size. Not different infrastructure providers.

I've tried to find any lingering mentions of "flavor" and I believe it should all be gone:
```bash
% fgrep -ri --exclude-dir kubespray --exclude-dir .git flavor . 
./docs/citycloud-snippets.md:OS_MASTERS_FLAVOR="" # UUID of master flavor
./docs/citycloud-snippets.md:OS_WORKER_FLAVOR="" # UUID of worker flavor
./docs/citycloud-snippets.md:      -e "s@^flavor_k8s_master = .*@flavor_k8s_master = \"${OS_MASTERS_FLAVOR}\"@" \
./docs/citycloud-snippets.md:      -e "s@^#\?flavor_k8s_node = .*@flavor_k8s_node = \"${OS_WORKER_FLAVOR}\"@" \
./WIP-CHANGELOG.md:- Renamed init argument flavor to cloud_provider.
```

Will update documentation if this gets approved.

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,16 +15,18 @@ Content:
 
     ```bash
     export CK8S_CONFIG_PATH=~/.ck8s/my-environment
-    ./bin/ck8s-kubespray init <prefix> <cloud_provider> [<SOPS fingerprint>]
+    ./bin/ck8s-kubespray init <prefix> [--cloud-provider cloud_provider] [--sops-fingerprint fingerprint]
     ```
 
     Arguments:
     * `prefix` will be used to differentiate this cluster from others in the same CK8S_CONFIG_PATH.
       For now you need to set this to `wc` or `sc` if you want to install compliantkubernetes apps on top afterwards, this restriction will be removed later.
-    * `cloud_provider` will determine some default values for a variety of config options.
+
+    Flags:
+    * `--cloud_provider` will determine some default values for a variety of config options, the environment variable `CK8S_CLOUD_PROVIDER` is used if unset.
       Supported options are `default`, `gcp`, `aws`, `vsphere`, and `openstack`.
-    * `SOPS fingerprint` is the gpg fingerprint that will be used for SOPS encryption.
-      You need to set this or the environment variable `CK8S_PGP_FP` the first time SOPS is used in your specified config path.
+    * `--sops-fingerprint` is the PGP fingerprint that will be used for SOPS encryption, the environment variable `CK8S_PGP_FP` is used if unset.
+      This flag is only required first time you configure SOPS in your specified config path.
 
 1. Edit the `inventory.ini` (found in your config path) to match the VMs (IP addresses and other settings that might be needed for your setup) that should be part of the cluster.
    Or if you have one created by a terraform script in `kubespray/contrib/terraform` you should use that one.
--- a/README.md
+++ b/README.md
@@ -15,13 +15,13 @@ Content:
 
     ```bash
     export CK8S_CONFIG_PATH=~/.ck8s/my-environment
-    ./bin/ck8s-kubespray init <prefix> <flavor> [<SOPS fingerprint>]
+    ./bin/ck8s-kubespray init <prefix> <cloud_provider> [<SOPS fingerprint>]
     ```
 
     Arguments:
     * `prefix` will be used to differentiate this cluster from others in the same CK8S_CONFIG_PATH.
       For now you need to set this to `wc` or `sc` if you want to install compliantkubernetes apps on top afterwards, this restriction will be removed later.
-    * `flavor` will determine some default values for a variety of config options.
+    * `cloud_provider` will determine some default values for a variety of config options.
       Supported options are `default`, `gcp`, `aws`, `vsphere`, and `openstack`.
     * `SOPS fingerprint` is the gpg fingerprint that will be used for SOPS encryption.
       You need to set this or the environment variable `CK8S_PGP_FP` the first time SOPS is used in your specified config path.
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -13,7 +13,7 @@ if [ $# -lt 1 ] || [ $# -gt 2 ]; then
     exit 1
 fi
 
-flavor=$1
+cloud_provider=$1
 if [ $# -eq 2 ]; then
     fingerprint=$2
 fi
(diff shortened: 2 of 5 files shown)


cristiklein (approved):
Awesome! Please check and update compliantkubernetes.io too.

Should we start a document to harmonize vocabulary, or is this a lonely mis-alignment between kubespray and apps?

Xartos (commented):
I'm not sure I'm aboard on moving it to a env variable.
I'd say it's way more transparent and explicit if we have it as an argument. Since this will only be used once and then forgotten it doesn't make sense to have it as a variable.

I'd rather see a move in the opposite direction, where we start using more arguments in the apps repo instead.

I totally agree with the namechange but I don't see the benefits of having it as a env instead of an argument

simonklb:
> I'm not sure I'm aboard on moving it to a env variable.
> I'd say it's way more transparent and explicit if we have it as an argument. Since this will only be used once and then forgotten it doesn't make sense to have it as a variable.

I agree with this, I'll take a look at how much work it would be to move to an argument in apps.

cristiklein:
@Xartos Can you elaborate how you would update the Kubespray config? I had something like this in mind:

```shellSession
$ export CK8S_CONFIG_PATH=~/ck-exoscale-config-ng/
$ . ${CK8S_CONFIG_PATH}/00-setup.sh # Contains CK8S_CLOUD_PROVIDER, CK8S_ENVIRONMENT_NAME and CK8S_FLAVOR
$ ck8s-kubespray init
$ git diff # operator reviews configuration changes
```

The issues I see with sending the cloud provider as a command line argument, is that you would need to "copy-paste" the provider from `$CK8S_CLOUD_PROVIDER` into the `ck8s-kubespray init` command line. This feels a bit more error-prone and "manual" than setting the flavor via env. Not sure if this is a strong enough argument. One could also argue that operators should be extra diligent when doing `init`. :smile: 

Let me know what you think.

simonklb:
> @Xartos Can you elaborate how you would update the Kubespray config? I had something like this in mind:
> 
> ```
> $ export CK8S_CONFIG_PATH=~/ck-exoscale-config-ng/
> $ . ${CK8S_CONFIG_PATH}/00-setup.sh # Contains CK8S_CLOUD_PROVIDER, CK8S_ENVIRONMENT_NAME and CK8S_FLAVOR
> $ ck8s-kubespray init
> $ git diff # operator reviews configuration changes
> ```
> 
> The issues I see with sending the cloud provider as a command line argument, is that you would need to "copy-paste" the provider from `$CK8S_CLOUD_PROVIDER` into the `ck8s-kubespray init` command line. This feels a bit more error-prone and "manual" than setting the flavor via env. Not sure if this is a strong enough argument. One could also argue that operators should be extra diligent when doing `init`. smile
> 
> Let me know what you think.

I guess this makes sense in the "re-init" case. It would be nice that you only had to specify the provider once and be done with it. If you run `init` again the tools should already know what provider you are using.

Xartos:
> @Xartos Can you elaborate how you would update the Kubespray config? I had something like this in mind:
> 
> ```
> $ export CK8S_CONFIG_PATH=~/ck-exoscale-config-ng/
> $ . ${CK8S_CONFIG_PATH}/00-setup.sh # Contains CK8S_CLOUD_PROVIDER, CK8S_ENVIRONMENT_NAME and CK8S_FLAVOR
> $ ck8s-kubespray init
> $ git diff # operator reviews configuration changes
> ```
> 
> The issues I see with sending the cloud provider as a command line argument, is that you would need to "copy-paste" the provider from `$CK8S_CLOUD_PROVIDER` into the `ck8s-kubespray init` command line. This feels a bit more error-prone and "manual" than setting the flavor via env. Not sure if this is a strong enough argument. One could also argue that operators should be extra diligent when doing `init`. smile
> 
> Let me know what you think.

I think that's a bit hidden, 

Let's say you're setting up a new cluster and you copy the `00-setup.sh` file then forget to change some variable in that file.
It's very hard to see what you're doing wrong then.

So in my mind it's a lot easier to directly recall what cluster you're setting up with the initial setup and just add them as arguments.

And for re-init, as simon called it, it would be really nice if that would be parsed from config. This would be really great for upgrades since the config is there already, and now you can basically write anything in those variables since they are often disregarded.

@simonklb I would really like to have clear separation for what's common settings and what's kubespray settings in that case. But if all the common settings were in `global.*` I think it would be okey. But we should have a clear API and responsibility for the different values in that case. E.g. if we update/add/remove a field, should it be changed by kubespray or apps or by both?

simonklb:
> @simonklb I would really like to have clear separation for what's common settings and what's kubespray settings in that case. But if all the common settings were in `global.*` I think it would be okey. But we should have a clear API and responsibility for the different values in that case. E.g. if we update/add/remove a field, should it be changed by kubespray or apps or by both?

Yea that's a tough question. You don't want to have two different values because you don't want to have kubespray be setup with provider X and apps be setup with provider Y. But you also don't want to have incompatibilities between them, so sharing a value would require some dependency which is also not nice.

It feels to me that eventually apps should not need to know what cloud provider it is running on but rather just look at what state the Kubernetes cluster is in. I'll investigate the reasons for why cloud provider needs to be specified in the configuration and see if there are some hard requirement for it.

simonklb:
Looks like the only reason why cloud provider is still left as an option in apps is for `citycloud->harbor->image chart storage->swift`. I believe we could get rid of cloud provider entirely from apps by making it configurable similarly to how we configure object storage. I'll take a stab at it.

simonklb:
I've now updated this to use flags instead of arguments with environment variables as fallback. PTAL!

simonklb:
Just FYI, I'm not against keeping cloud provider as an argument instead of flag since it's always required as long as the CK8S_CLOUD_PROVIDER environment variable isn't causing an error.

The only reason I went with a flag is because someone might be setting CK8S_CLOUD_PROVIDER and expecting it to be used.

simonklb:
As I'm leaving for vacation now feel free to merge/update this in my place or leave this hanging until I get back!

Xartos (approved):
LGTM!

README.md
```diff
@@ -15,16 +15,18 @@ Content:
...
     ```
 
     Arguments:
     * `prefix` will be used to differentiate this cluster from others in the same CK8S_CONFIG_PATH.
       For now you need to set this to `wc` or `sc` if you want to install compliantkubernetes apps on top afterwards, this restriction will be removed later.
-    * `flavor` will determine some default values for a variety of config options.
+
+    Flags:
+    * `--cloud_provider` will determine some default values for a variety of config options, the environment variable `CK8S_CLOUD_PROVIDER` is used if unset.
       Supported options are `default`, `gcp`, `aws`, `vsphere`, and `openstack`.
```
viktor-f, line 27: This is not correct anymore.
```suggestion
      Supported options are `citycloud`, `safespring`, `gcp`, `aws`, `vsphere`, and `exoscale`.
```

bin/init.bash
```diff
@@ -8,44 +8,52 @@
...
+
+        citycloud|safespring)
+            config_type="openstack"
+            ;;
+
+        *)
+            log_error "ERROR: Unsupported cloud provider: ${cloud_provider}"
+            exit 1
+            ;;
+    esac
```
viktor-f, line 50: This does not support exoscale. Should there be a way to generate a "generic" config as well?
```suggestion
if [ -n "${cloud_provider:-}" ]; then
    case "${cloud_provider:-}" in
        aws|gcp|vsphere)
            config_type="${cloud_provider}"
            ;;

        citycloud|safespring)
            config_type="openstack"
            ;;
            
        exoscale)
            ;;

        *)
            log_error "ERROR: Unsupported cloud provider: ${cloud_provider}"
            exit 1
            ;;
    esac
```

viktor-f (changes requested):

OlleLarsson:
What do we do with this PR? Close?

viktor-f:
> What do we do with this PR? Close?

Yeah it's time.

PR #105, opened 2021-08-06
apps: added resources requests to rook containers

crssnd:
**What this PR does / why we need it**: missing resource requests for containers 

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #580

- the values for the resource requests cpu and memory are based on the containers consumption from my exo dev cluster (24h period)
- I didn't find a solution for adding resource requests to kube-proxy, kube-apiserver, kube-scheduler and kube-controller. The manifests from kubernetes do not contain this fields and cannot be added or costumized via kubeadm. We can include some manual commands/scripts to add this after the cluster is up and running, but I do not think we want to.
- the update to rook containers went well in my cluster, but we should test more

sc_req

wc-req

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -8,5 +8,6 @@
 ### Added
 
 - Added new command to reboot nodes in a cluster if necessary
+- resource requests for rook containers [#105](https://github.com/elastisys/compliantkubernetes-kubespray/pull/105)
 
 ### Removed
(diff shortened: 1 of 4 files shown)


OlleLarsson (approved):
LGTM

OlleLarsson:
I think you can squash the commits into a single one.

viktor-f (approved):
LGTM, and yes please squash the commits before merging.

PR #114, opened 2021-09-16
Reset changelog for release v2.17.0-ck8s1

crssnd:
**What this PR does / why we need it**: needed for step 3 from the release process


viktor-f:
This is the wrong version number, see my comments on slack. In short we are not updating kubespray so the new version should be `v2.16.0-ck8s2`

PR #118, opened 2021-09-29
config: Update Openstack config to use Cinder by default

### LINKED ISSUE #110: [1] Configure cinder-csi storageclass from ansible
**Is your feature request related to a problem?**
When deploying on Openstack with cinder enabled, the storageclass `cinder-csi` is not default nor is volume expansion enabled.

Enabling volume expansion and setting it to be the deafult storageclass is done by default on the supportet clouds.
See [kubespray](https://github.com/kubernetes-sigs/kubespray/blob/v2.16.0/roles/kubernetes-apps/persistent_volumes/cinder-csi/defaults/main.yml#L2-L7) for variables.

**Additional context**

**Definition of done:**
The cinder-csi is the deafult storageclass and volume expansion is enabled.

aarnq:
This sets Cinder CSI as the default storage provider on Openstack while enabling volume expansion on new clusters.

fixes #110 

https://github.com/elastisys/compliantkubernetes/pull/177

**Checklist:**

- [x] Updated the [public facing documentation](https://github.com/elastisys/compliantkubernetes)
  - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -10,5 +10,6 @@
 
 - Added new command to reboot nodes in a cluster if necessary
 - resource requests for rook containers [#105](https://github.com/elastisys/compliantkubernetes-kubespray/pull/105)
+- Added configuration for Openstack to use Cinder CSI by default with volume expansion enabled.
 
 ### Removed
--- a/config/openstack/group_vars/k8s_cluster/ck8s-k8s-cluster-openstack.yaml
+++ b/config/openstack/group_vars/k8s_cluster/ck8s-k8s-cluster-openstack.yaml
@@ -8,9 +8,17 @@ external_openstack_cloud_controller_extra_args:
   # Must be different for every cluster in the same openstack project
   cluster-name: "set-me"
 
-# cinder_csi_enabled: true
-# persistent_volumes_enabled: true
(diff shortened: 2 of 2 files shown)


aarnq:
I've changed volume expansion to be opt in, as it creates some issues for me on SafeSpring with volumes getting stuck.

Xartos (approved):
:tada:

crssnd:
> I've changed volume expansion to be opt in, as it creates some issues for me on SafeSpring with volumes getting stuck.

it doesn't work at safe spring by default? maybe you can add a comment about this.

please do not forget to squash your commits

aarnq:
> > I've changed volume expansion to be opt in, as it creates some issues for me on SafeSpring with volumes getting stuck.
> 
> it doesn't work at safe spring by default? maybe you can add a comment about this.

Now it kind of does after they fixed issues using the API from within the cluster. I will update it.

In SafeSpring the volume PV and PVC may present themselves as resized, but the node and Pod might not see it. If the node can't see it then the Pod must be restarted, if the node _can_ see but not the Pod then you have to resize the filesystem from the node using (for ext4) `resize2fs`.

I will see if I can add that to the ops repo.

> please do not forget to squash your commits

:+1:

crssnd (approved):

PR #119, opened 2021-10-05
docs: Added known issue for rebooting node script

### LINKED ISSUE #112: [1] reboot-nodes script throws an error
**Describe the bug**

When running the reboot-nodes script it sometimes fails when waiting for a node to come back up. The error is

fatal: [staging-sc-k8s-master-1]: FAILED! => {"msg": "The conditional check 'kubelet_heartbeat.rc == 0' failed. The error was: error while evaluating conditional (kubelet_heartbeat.rc == 0): 'dict object' has no attribute 'rc'"}

If the script is executed again it usually works for that node, but might stop on the next, requiring multiple re-runs to reboot all nodes.

**To Reproduce**
Run ./bin/ck8s-kubespray reboot-nodes [sc|wc]

**Expected behavior**
All nodes are rebooted without error

**Screenshots**

- Compliant kubernetes apps version [e.g. v0.8.0]
- compliantkubernetes-kubespray 

**Additional context**

robinAwallace:

Added a note that there is a known issue for the reboot script. I could not recreate the bug, i tried multiple ways to make the step fail but the error was never the same. 

The error occurs if the `kubelet_heartbeat` is not registered in the playbook, which is unclear why it happens. 

I also increased the number of retries, as it may be because of to few retries.

fixes #112 

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,11 @@ You can reboot all nodes that wants to restart (usually to finish installing new
 
 If you set `--extra-vars manual_prompt=true` then you get a manual prompt before each reboot so you can stop the playbook if you want.
 
+### Known issues
+
+- The script may fail with the message `error while evaluating conditional (kubelet_heartbeat.rc == 0): 'dict object' has no attribute 'rc'`
+  - In such situations just rerun the script. It will skip the check for that node, so check that it is up and running manually afterwards.
+
 ## Running other kubespray playbooks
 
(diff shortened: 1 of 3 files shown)


cristiklein (approved):
Thanks for documenting this!

Xartos (approved):

PR #133, opened 2021-11-03
apps:rook add provisioner and plugin tolerations for tainted nodes

lucianvlad:
**What this PR does / why we need it**:  I have added in commented lines the "provisionerTolerations" and "pluginTolerations". These are needed for Exoscale and other cloud providers where we use rook-ceph. It is needed to be able to schedule the csi-rbdplugin pods on nodes that are tainted and be able to mount volumes on those tainted nodes.
Without this, the tainted nodes will not have the csi-rbdplugin pods running on them, and without that the nodes will fail to mount volumes with the following error:

_FailedMount  9m25s (x461 over 15h)  kubelet  MountVolume.MountDevice failed for volume "pvc-9e7ec0da-d52d-46b8-bd74-e5262286cb0d" : kubernetes.io/csi: attacher.MountDevice failed to create newCsiDriverClient: driver name rook-ceph.rbd.csi.ceph.com not found in the list of registered CSI drivers_

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/rook/operator-values.yaml
+++ b/rook/operator-values.yaml
@@ -1,5 +1,13 @@
 csi:
   enableCephfsDriver: false
+  #provisionerTolerations:
+  #  - key: "nodeType"
+  #    operator: Exists
+  #    effect: NoSchedule
+  #pluginTolerations:
+  #  - key: "nodeType"
+  #    operator: Exists
+  #    effect: NoSchedule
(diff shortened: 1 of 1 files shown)


Xartos (approved):
Super!

viktor-f (approved):

PR #147, opened 2021-12-21
bin: add prompt and checks for openstack credentials

### LINKED ISSUE #146: 1 Ensuring that the operator is working on the correct environment

When working on multiple openstack environments it is possible to forget to parse the correct credentials. This is a problem for the openstack controller if the wrong credentials are sourced so it will receive the wrong credentials and will start to fail.  

There are two simple options:

1. Prompting the operator with the current variables and ask if they are correct.
2. Adding a test that ensure that the cloud controller is up and operational. Test to provision a pvc if supported by the infra provider...

**Additional context**

**Definition of done:**
When could this feature be considered done.

When there is a prompt to the user or a test in place ensuring that the cloud controller has the correct config.

viktor-f:
**What this PR does / why we need it**: Adds some checks for openstack credentials and a prompt for the user to hopefully catch when a user would accidentally use the wrong credentials.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #146 

Some examples:

No credentials set:
no-creds

`OS_USERNAME` and `OS_PASSWORD` are set:
os-user

`OS_APPLICATION_CREDENTIAL_SECRET` is missing:
no-appsecret

Both openstack user and openstack application credentials are set:
both-creds

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/bin/apply.bash
+++ b/bin/apply.bash
@@ -11,6 +11,8 @@ here="$(dirname "$(readlink -f "$0")")"
 # shellcheck source=bin/common.bash
 source "${here}/common.bash"
 
+check_openstack_credentials
+
 log_info "Creating kubernetes cluster using kubespray"
 pushd "${kubespray_path}"
 
--- a/bin/common.bash
+++ b/bin/common.bash
@@ -46,6 +46,10 @@ log_info() {
     echo -e "[\e[34mck8s\e[0m] ${*}" 1>&2
 }
 
+log_info_no_newline() {
+    echo -e -n "[\e[34mck8s\e[0m] ${*}" 1>&2
+}
+
 log_warning() {
     echo -e "[\e[33mck8s\e[0m] ${*}" 1>&2
 }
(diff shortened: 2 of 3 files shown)


bin/common.bash
```diff
@@ -175,3 +179,51 @@ sops_exec_file_no_fifo() {
...
+# Checks the current openstack env variables to see if anything is missing.
+# The user is then show what is set and prompted if they want to proceed or not.
+# The user can still proceed if nothing is set, to allow for other types of cloud providers.
+check_openstack_credentials() {
+    log_info "Checking for openstack user or openstack application credentials"
+
+    if [ -n "${OS_USERNAME:-}" ] && [ -n "${OS_APPLICATION_CREDENTIAL_NAME:-}" ]; then
+        log_warning "Warning: Both OS_USERNAME and OS_APPLICATION_CREDENTIAL_NAME is set."
+        log_warning "Will default to using Application credentials. You can unset application credentials to use openstack user instead."
+    fi
```
viktor-f, line 192: Should I add this to be 100% sure what credentials are used?
```suggestion
    if [ -n "${OS_USERNAME:-}" ] && [ -n "${OS_APPLICATION_CREDENTIAL_NAME:-}" ]; then
        log_warning "Warning: Both OS_USERNAME and OS_APPLICATION_CREDENTIAL_NAME is set."
        log_warning "Will default to using Application credentials. You can unset application credentials to use openstack user instead."
        unset OS_USERNAME
        unset OS_PASSWORD
    fi
```

crssnd: here I would just exit and let the user decide what credentials he/she wants to use
maybe just print the unset commands for easy copy/paste

viktor-f: Good idea. That might be the best solution.

crssnd (approved):

bin/apply.bash
```diff
@@ -11,6 +11,8 @@ here="$(dirname "$(readlink -f "$0")")"
 # shellcheck source=bin/common.bash
 source "${here}/common.bash"
 
+check_openstack_credentials
```
OlleLarsson, line 14: Do we want this in the run-playbook as well?

OlleLarsson (approved):

bin/apply.bash
OlleLarsson: It was there :)

bin/common.bash
viktor-f: Fixed now.

PR #164, opened 2022-02-28
Rook fixes

cristiklein:

The rook-ceph-toolsbox image change and CrashLoops if created with the wrong Deployment. Pulling the latest version.

Also, the deploy script is not idempotent and may fail with:

```
error: 'owner' already has a value (operator), and --overwrite is false
```

N/A

N/A

**Checklist:**

  - [X] is completely transparent, will not impact the workload in any way.
  - [ ] will break the cluster.

### DIFF
--- a/rook/toolbox-deploy.yaml
+++ b/rook/toolbox-deploy.yaml
@@ -1,8 +1,13 @@
+# This file was originally downloaded from https://github.com/rook/rook/blob/c890710b635e2cd2ae64c08bac0621d89e79c0fc/deploy/examples/toolbox.yaml
+#
+# The following changes were performed:
+# - use stable container image, i.e., 'v1.8.6' instead of master
+# - add resource limits
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: rook-ceph-tools
-  namespace: rook-ceph # namespace:cluster
+  namespace: rook-ceph
   labels:
     app: rook-ceph-tools
 spec:
@@ -18,7 +23,7 @@ spec:
       dnsPolicy: ClusterFirstWithHostNet
       containers:
         - name: rook-ceph-tools
-          image: rook/ceph:master
+          image: rook/ceph:v1.8.6
           command: ["/bin/bash"]
           args: ["-m", "-c", "/usr/local/bin/toolbox.sh"]
           imagePullPolicy: IfNotPresent
@@ -43,6 +48,10 @@ spec:
               name: ceph-config
             - name: mon-endpoint-volume
               mountPath: /etc/rook
+          resources:
+            requests:
+             cpu: "5m"
+             memory: "10Mi"
       volumes:
         - name: mon-endpoint-volume
           configMap:
--- a/rook/toolbox-deploy.yaml
+++ b/rook/toolbox-deploy.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: rook-ceph-tools
+  namespace: rook-ceph # namespace:cluster
   labels:
     app: rook-ceph-tools
 spec:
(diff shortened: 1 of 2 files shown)


rook/toolbox-deploy.yaml
```diff
@@ -17,10 +18,15 @@ spec:
       dnsPolicy: ClusterFirstWithHostNet
       containers:
         - name: rook-ceph-tools
-          image: rook/ceph:v1.5.3
-          command: ["/tini"]
-          args: ["-g", "--", "/usr/local/bin/toolbox.sh"]
+          image: rook/ceph:master
```
Xartos, line 21: Isn't `master` basically same as `latest`? I wouldn't like that

rook/toolbox-deploy.yaml
```diff
@@ -17,10 +18,15 @@ spec:
...
+          command: ["/bin/bash"]
```
Xartos, line 22: Any reason to this change?

rook/toolbox-deploy.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: rook-ceph-tools
+  namespace: rook-ceph # namespace:cluster
```
Xartos, line 5:
```suggestion
  namespace: rook-ceph
```

rook/toolbox-deploy.yaml
```diff
@@ -37,10 +43,6 @@ spec:
               name: ceph-config
             - name: mon-endpoint-volume
               mountPath: /etc/rook
-          resources:
-            requests:
-             cpu: "5m"
-             memory: "10Mi"
```
Xartos, line 43: Why remove resource request? This doesn't follow our best practice. I would rather add limits to it as well.

Xartos (changes requested):

rook/toolbox-deploy.yaml
cristiklein: I was as surprised as you are, but seems like Rook really insists on `master`:
- https://github.com/rook/rook/blob/master/deploy/examples/toolbox.yaml#L21
- https://hub.docker.com/r/rook/ceph/tags?page=1&name=latest
- https://hub.docker.com/r/rook/ceph/tags?page=1&name=master

rook/toolbox-deploy.yaml
cristiklein: CrashLoop. :smile: The newer images do not include `/tini`. We could use an older version of `rook/ceph` in the toolbox, but since this is not on the "critical path" rather it's for debugging, I think that giving the operator the latest and greatest image is nicer.

rook/toolbox-deploy.yaml
cristiklein: I simply copied the file from here: https://github.com/rook/rook/blob/master/deploy/examples/toolbox.yaml#L5

Unsure why they added the comment there, but since I didn't change anything else, it felt weird to change only this.

rook/toolbox-deploy.yaml
cristiklein: Okay, so you argue that we need to depart a bit from the official toolbox manifest, right?

Or should we upstream the `resources`?

Xartos: But otherwise I don't see an issue with deviating in this case since this is just a single manifest tool. So I don't see any burden to manage these small changes.

cristiklein:
@Xartos I addressed all your comments, specifically:

- made container image stable
- added resource requests
- removed comment after namespace

I also added a comment to clearly document origin of file and what changes we made. PTAL.

Xartos (approved):
LGTM!

OlleLarsson (approved):

PR #187, opened 2022-05-02
bin: add remove-node command

### LINKED ISSUE #175: [1]Ensure node variable is set when running remove-node playbook
When running the remove-node.yml playbook you need to set the variable `node` to limit the execution to that host.
If node is not set it will run though all of the worker nodes.

It can happen due to a mistake, spelling or otherwise, that the node variable is not set properly whereby kubespray tries to remove all worker nodes. There is a confirm box when running this playbook, but when it is run for all worker nodes it does not say this in any super obvious way, therefore it can be missed that it is targeting all nodes.

I propose that we create a command for removing nodes to prevent the node variable from being unset.

`bin/ck8s-kubespray remove-node {wc,sc} <node_name>`

Keep using the run-playbook command

**Additional context**
Mistakes happens to everyone, therefore we should make it harder for them to occur whenever we can 😃 

**Definition of done:**
When there is a command that ensures that the node variable is set and passed to the remove-node playbook

OlleLarsson:

fixes #175 

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,0 +1,2 @@
+### Added
+- Added `remove-node` command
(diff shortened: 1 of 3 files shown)


crssnd (approved):
please update or create an issue to update the internal docs with this new script

viktor-f (approved):

PR #216, opened 2022-09-16
rook: add elastisys toleration

crssnd:
**What this PR does / why we need it**: to add the elastisys tolerations

- not sure if we want the tolerations here or not, but they are needed for devbox

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/rook/cluster.yaml
+++ b/rook/cluster.yaml
@@ -21,6 +21,17 @@ spec:
       enabled: true
   dashboard:
     enabled: false
+  placement:
+    all:
+      tolerations:
+      - effect: NoSchedule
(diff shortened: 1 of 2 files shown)


rook/cluster.yaml
```diff
@@ -21,6 +21,17 @@ spec:
       enabled: true
   dashboard:
     enabled: false
+  placement:
+    all:
+      tolerations:
+      - effect: NoSchedule
+        key: elastisys.io/node-type
```
robinAwallace, line 28: Do we want to have elastisys specific configurations here?

PR #242, opened 2022-11-30
Remove rest of docker files

robinAwallace:

Remove the rest of docker files after migrating to containerd

**Which issue this PR fixes**: fixes #

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/migration/v2.18.0-ck8s1-v2.19.x-ck8s1/migrate-to-containerd.yml
+++ b/migration/v2.18.0-ck8s1-v2.19.x-ck8s1/migrate-to-containerd.yml
@@ -55,6 +55,18 @@
           ansible.builtin.debug:
             var: clusterymloutput
 
+        - name: Delete remaining docker files
+          file:
+            state: absent
+            path: "{{ item }}"
+          with_items:
+            - '/var/lib/docker'
+            - '/var/lib/dockershim'
+            - '/etc/systemd/system/docker.service.d'
+            - '/etc/systemd/system/docker.service'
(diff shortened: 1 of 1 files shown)


OlleLarsson (commented):
Should we also do `systemctl daemon-reload` after removing the unit files?

robinAwallace:
> Should we also do `systemctl daemon-reload` after removing the unit files?

You could, but as the next steps include reboot it will happen automatically. Thats why i skipped it

OlleLarsson (approved):

OlleLarsson:
> > Should we also do `systemctl daemon-reload` after removing the unit files?
> 
> You could, but as the next steps include reboot it will happen automatically. Thats why i skipped it

I see

migration/v2.18.0-ck8s1-v2.19.x-ck8s1/migrate-to-containerd.yml
```diff
@@ -55,6 +55,18 @@
           ansible.builtin.debug:
             var: clusterymloutput
 
+        - name: Delete remaining docker files
```
Xartos, line 58: Do we want to delete them in this step? It would be nicer to do it after we've migrated all so that we can revert back if in the middle something goes wrong

Xartos: But at the same time if we just re-add docker it will probably set itself up again so maybe this is a non-issue

PR #260, opened 2023-02-17
Updated kubespray to support network id

Xartos:

For metallb we need to have support to override network ID on individual nodes which was missed when we did the last patch. So this would fix that. But this is part of the kubespray 2.21 release so we don't need to fix it on main

**Checklist:**

  - [ ] will break the cluster.


robinAwallace (approved):

davidumea (approved):

davidumea:
So you're creating a -ck8s4 patch with this?

Xartos:
> So you're creating a -ck8s4 patch with this?

Yes, that was the plan

PR #262, opened 2023-02-27
bin: changed to inventory hostnames in reboot script

### LINKED ISSUE #261: Reboot_nodes do not work with override_system_hostname: false
**Describe the bug**
Kubespray can be configured to not change the actual hostname of the machines with the setting:
`override_system_hostname: false`

The reboot-nodes playbook seems to use the actual hostname of the machines, and not what is specified in the inventory file. So if you don't override hostnames, and use different node names in the inventory file, this script will not work.

**To Reproduce**

1. Configure kubespray with `override_system_hostname: false`
1. Use node names in `inventory.ini` that are different to the real hostnames
1. Run `ck8s-kubespray reboot-nodes`
1. It should fail 

**Expected behavior**
Kubespray should be able to run playbooks against nodes even with `override_system_hostname: false`

- Compliant kubernetes kubespray version v2.20.0-ck8s3

**Additional context**
I'm not sure if this is also an issue for any other playbook.

Elias-elastisys:
Changed reboot script to use the hostnames specified in the inventory file instead of the actual hostnames of the machines. This allows for the `override_system_hostname: false` setting to be used.

**Which issue this PR fixes**: fixes #261

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/playbooks/reboot_nodes.yml
+++ b/playbooks/reboot_nodes.yml
@@ -25,15 +25,8 @@
       when: "manual_prompt | default(false) | bool and 'reboot_required_file' in hostvars[item] and hostvars[item]['reboot_required_file'].stat.exists"
       with_items: "{{ play_hosts }}"
 
-    - name: hostname
-      shell: echo $HOSTNAME
-      args:
-        executable: /bin/bash
-      register: hostname
-      when: reboot_required_file.stat.exists
-
     - name: drain node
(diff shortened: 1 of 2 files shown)


playbooks/reboot_nodes.yml
```diff
@@ -33,7 +33,7 @@
       when: reboot_required_file.stat.exists
```
robinAwallace, line 33: As this hostname is not used anymore you can remove it.

robinAwallace (approved):
LGTM :+1:

Xartos (approved):

PR #263, opened 2023-02-28
Qa 2.21.0

davidumea:

**Checklist:**

  - [X] requires running a migration script.
  - [ ] will break the cluster.


migration/v2.20.0-ck8sx-v2.21.0-ck8s1/upgrade-cluster.md
```diff
@@ -0,0 +1,128 @@
...
+        mv terraform-temp.tfstate terraform.tfstate
+        popd > /dev/null
+    done
+    ```
+
+1. Lastly apply the new state to update the modules (there should be no diff in the terraform output here)
+
+    ```bash
+    ./migration/v2.20.0-ck8sx-v2.21.0-ck8s1/apply-new-state.sh
+    ```
```
aarnq, line 109: Are these only for openstack?

davidumea: Yes, I should add an extra note there :+1:

aarnq (approved):

migration/v2.20.0-ck8sx-v2.21.0-ck8s1/migrate-terraform-openstack-user-data.sh
```diff
@@ -0,0 +1,33 @@
...
+    if ! terraform -chdir="${openstack_terraform_dir}" plan -var-file="${ck8s_kubespray_config_path}/cluster.tfvars" -state="${ck8s_kubespray_config_path}/terraform-temp.tfstate" -detailed-exitcode; then
+        echo "Terraform found changes for $CLUSTER-cluster, review the changes."
+        echo "Continuing here will not apply anything, it will just create a temporary state file."
+        read -rp "Continue? [y/N] " val
+        case "${val}" in
+            [Yy]) ;;
+            *) exit ;;
+        esac
+    fi
+    popd || return
```
Xartos, line 32: Should we maybe print the command to run if it succeeds? Something like:

```suggestion
    popd || return
    echo "If you want to apply, run the following:"
    terraform -chdir="${openstack_terraform_dir}" apply -var-file="${ck8s_kubespray_config_path}/cluster.tfvars" -state="${ck8s_kubespray_config_path}/terraform-temp.tfstate" -detailed-exitcode
```

migration/v2.20.0-ck8sx-v2.21.0-ck8s1/upgrade-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+        +podsecuritypolicy_enabled: false
+        ```
+
+    1. Remove `PodSecurityPolicy` from the `kube_apiserver_enable_admission_plugins` list in both `sc-config/group_vars/k8s_cluster/ck8s-k8s-cluster.yaml` and `wc-config/group_vars/k8s_cluster/ck8s-k8s-cluster.yaml`
+
+        ```diff
+         kube_apiserver_enable_admission_plugins:
+        -  - "PodSecurityPolicy"
+           - "NamespaceLifecycle"
+        ```
```
Xartos, line 68: I realized that I would also like to add
```suggestion
        ```

1. Set shutdown manager to valid values for Ubuntu

    ```bash
    kubelet_shutdown_grace_period: 30s
    kubelet_shutdown_grace_period_critical_pods: 10s
    ```
```

migration/v2.20.0-ck8sx-v2.21.0-ck8s1/migrate-terraform-openstack-user-data.sh
davidumea: It's being applied later with another script, I felt like it made more sense to apply the `terraform.tfstate` file instead of the `-temp` file since they change a bit after apply, even though they will look the same in the end.

robinAwallace (approved):

migration/v2.20.0-ck8sx-v2.21.0-ck8s1/migrate-terraform-openstack-user-data.sh
Xartos: Right, I see that now

Xartos (approved):

PR #266, opened 2023-03-20
config: initial rework of audit log policy

### LINKED ISSUE #253: Reduce kubernetes audit log verbosity
.. to satisfy only the needs for HIPAA/PDL.

https://github.com/elastisys/compliantkubernetes-kubespray/blob/main/config/common/group_vars/k8s_cluster/ck8s-k8s-cluster.yaml#L47

Kubernetes only outputs audit logs relevant for HIPAA/PDL.

Leave current logging verbosity and then add another template for when HIPAA/PDL compliance is needed.

**Definition of done:**
Group vars for audit logging is minimal but enough to satisfy the requirements of HIPAA/PDL.

OlleLarsson:
Initial rework of the Kubernetes audit log policy.
The policy is based on the the one that is generated by GKE wich also happens to be the default one [included by Kubespray](https://github.com/kubernetes-sigs/kubespray/blob/master/roles/kubernetes/control-plane/templates/apiserver-audit-policy.yaml.j2).
The default policy has been modified slightly by for example removing some rule that no longer is relevant.
The default log level was chosen to be RequestResponse.

The approach taken now is to reduce the number of logged events by mainly dropping read requests.

Right now the policy will be applied for both WC and SC, in the future it makes sense to have one for each type of cluster.
Note that the policy was reworked from the perspective of the WORKLOAD CLUSTER, therefore I will create follow up issues not only to modify  it to target components in the service cluster, but also to go into a bit more detail one each application/component.  

fixes #253 

**Public facing documentation PR**
https://github.com/elastisys/compliantkubernetes/issues/532

**Checklist:**

  - [x] requires running a migration script.
  - [ ] will break the cluster.


cristiklein (approved):
Awesome work, thank you!

crssnd (approved):
took a quick look and looks good

viktor-f (approved):

PR #269, opened 2023-04-19
rook-ceph: Add PSA labels and disable PSP

lunkan93:
 Add PSA labels and disable Kubernetes PSPs for rook-ceph
**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: Part of fix for [#1451](https://github.com/elastisys/compliantkubernetes-apps/issues/1451)

**Checklist:**

  - [ ] will break the cluster.

### DIFF
--- a/rook/deploy-rook.sh
+++ b/rook/deploy-rook.sh
@@ -14,6 +14,9 @@ chart_version="v1.10.5"
 # Install rook operator
 kubectl create namespace "${namespace}" --dry-run -o yaml | kubectl apply -f -
 kubectl label namespace "${namespace}" owner=operator --overwrite
+kubectl label namespace "${namespace}" pod-security.kubernetes.io/audit=privileged
+kubectl label namespace "${namespace}" pod-security.kubernetes.io/enforce=privileged
+kubectl label namespace "${namespace}" pod-security.kubernetes.io/warn=privileged
 helm upgrade --install --namespace "${namespace}" "${release_name}" "${chart}" \
   --version "${chart_version}" --values "${here}/operator-values.yaml" --wait
 
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,6 +1,8 @@
 ### Changed
 
 - Updated the Kubernetes audit log policy file
+- Added PSA labels for rook-ceph
+- Disabled Kubernetes PSPs for rook-ceph
 
 ### Fixed
 - Reboot scripts uses inventory hostnames instead of machine hostnames
--- a/migration/v2.20.0-ck8sx-v2.21.0-ck8s1/upgrade-cluster.md
+++ b/migration/v2.20.0-ck8sx-v2.21.0-ck8s1/upgrade-cluster.md
@@ -50,6 +50,10 @@
     ntp_timezone: "Etc/UTC"
     ```
 
+1. **If the environment is running `rook-ceph`**
+
+    Set `pspEnable` to `false` in `rook/operator-values.yaml`
+
 1. Disable Pod Security Policies
 
     1. Change the following variable in both `sc-config/group_vars/k8s_cluster/ck8s-k8s-cluster.yaml` and `wc-config/group_vars/k8s_cluster/ck8s-k8s-cluster.yaml`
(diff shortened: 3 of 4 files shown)


rook/deploy-rook.sh
```diff
@@ -14,6 +14,9 @@ chart_version="v1.10.5"
 # Install rook operator
 kubectl create namespace "${namespace}" --dry-run -o yaml | kubectl apply -f -
 kubectl label namespace "${namespace}" owner=operator --overwrite
+kubectl label namespace "${namespace}" pod-security.kubernetes.io/audit=privileged --overwrite
```
robinAwallace, line 17: I do not think you need the overwrite?
```suggestion
kubectl label namespace "${namespace}" pod-security.kubernetes.io/audit=privileged
```

robinAwallace (approved):

Xartos (commented):
Do we need some migration guide on this as well? Like when upgrading to 2.21 I guess we need to manually update the namespace for clusters that run rook?

lunkan93:
> Do we need some migration guide on this as well? Like when upgrading to 2.21 I guess we need to manually update the namespace for clusters that run rook?

Should this be added for migrating to `v2.21.0-ck8s2` or `v2.22.0-ck8s1`?

Xartos:
To `v2.21.0-ck8s2` since we aren't changing the kubespray version, only changing stuff on our side, it's the patch version.

migration/v2.21.0-ck8s1-v2.21.0-ck8s2/upgrade-cluster.md
```diff
@@ -0,0 +1,70 @@
...
+These steps will cause disruptions in the environment.
+
+1. Upgrade the cluster to a new kubernetes version:
+
+    ```bash
+    ./bin/ck8s-kubespray run-playbook sc upgrade-cluster.yml -b --skip-tags=download
+    ./bin/ck8s-kubespray run-playbook wc upgrade-cluster.yml -b --skip-tags=download
+    ```
+
+1. If the environment is running `rook-ceph`, apply PSA labels to the namespace:
```
Xartos, line 45: Will this work? If you upgrade the cluster (hence you will loose PSP support I assume) and then before you've applied these labels, will it not get the `baseline` PSA?

Wouldn't it be better to first add these labels to the namespace and then upgrade instead?

migration/v2.21.0-ck8s1-v2.21.0-ck8s2/upgrade-cluster.md
```diff
@@ -0,0 +1,70 @@
...
+
+    ```bash
+    kubectl label namespace rook-ceph pod-security.kubernetes.io/audit=privileged
+    kubectl label namespace rook-ceph pod-security.kubernetes.io/enforce=privileged
+    kubectl label namespace rook-ceph pod-security.kubernetes.io/warn=privileged
+    ```
```
Xartos, line 51: Do we need to add a step to also apply the new operator values? Or will the old PSPs be automatically removed?

lunkan93: Should this be done before or after the upgrade? :thinking:

Xartos: Good question, I guess if we've enabled the PSA labels it will not break if we remove the PSPs before the upgrade. However I'd like to get this tested before we go with that

lunkan93: I'm thinking these migration steps should actually be performed when migrating from 2.20 to 2.21, no matter what patch version you're migrating to, in addition to the [steps](https://github.com/elastisys/compliantkubernetes-kubespray/blob/v2.21.0-ck8s1/migration/v2.20.0-ck8sx-v2.21.0-ck8s1/upgrade-cluster.md) that are already there. The steps are not specifically tied to this PR and don't really make sense as I have added them now, as if you were migrating from `v2.21.0-ck8s1` to `v2.21.0-ck8s2` you would most likely already have run into issues if you didn't perform these steps when you upgraded to `v2.21.0-ck8s1` in the first place (In environments running `rook-ceph`).

Xartos: Yes, I totally agree. Let's add this to the 2.20 -> 2.21 upgrade instead

Xartos (commented):
One minor nit, otherwise :+1:

Xartos (approved):

PR #271, opened 2023-05-11
Expose version as config and change migration process to automate updating version

Ajarmar:
**What this PR does / why we need it**: This PR does a few things:

- Adds a new config file with a value `ck8sKubesprayVersion`, corresponding to the `compliantkubernetes-kubespray` version.
- Adds an "upgrade prepare" command that works similarly to how it works in `compliantkubernetes-apps`; updates the config to prepare for a version upgrade. (This includes automatic update of the version config value.)
- Updates existing migration guides for recent versions to use this new prepare command, as well as adds a migration guide for a new ck8s patch version for v2.20 that just adds this new config value. (I'll create this patch release after this PR is merged.)

**Special notes for reviewer**: This is a pretty big change to the migration process, for a PR that wasn't primarily about changing the migration process. However, since we decided that we wanted a "prepare" command, I felt that it was best to make it as similar to how it works in `compliantkubernetes-apps` as possible. It's basically a slimmed-down version of the apps migration process (no "upgrade apply" command).

**Checklist:**

  - [ ] will break the cluster.


bin/ck8s-kubespray
```diff
@@ -23,6 +23,8 @@ usage() {
     echo "      args: <prefix> [--extra-vars manual_prompt=true] [<options>]" 1>&2
```
davidumea, line 23: Shouldn't we have something in this binary to check if `ck8sKubesprayVersion` on your kubespray branch matches the one set in `CK8S_CONFIG_PATH`? I was expecting to see something like this

```
[ck8s] ERROR: Version mismatch. Run init to update config.
[ck8s] Config version: v2.20.0-ck8s3
[ck8s] CK8S version: 04b39733c17cd76370addae531e53db4616fe8a1
```
when running this command on mismatching versions
```
./bin/ck8s-kubespray run-playbook wc facts.yml
```

Ajarmar: Yeah, I can add something like that. I don't think `ck8s-kubespray init` is an idempotent command though, it's not really built to be run every time you do an upgrade like in apps. I could change that though

Ajarmar: I pushed a commit that adds this functionality, but I decided to not change how `init` works because making it so that it updates the version in the config may make the user expect it to work like apps init where it also makes other changes to the config. (Changing kubespray init to work just like apps init seems like it would be complicated because of how kubespray config differs from apps config, particularly the lack of default/override split.)

robinAwallace (commented):
LGTM, i would like to see a demo before approving :+1:

migration/v2.20.0-ck8s4/upgrade-cluster.md
```diff
@@ -0,0 +1,34 @@
...
+1. Run the following to create the new config file for both sc and wc.
+
+    ```console
+    export CK8S_KUBESPRAY_PATH=/path/to/compliantkubernetes-kubespray
+
+    for cluster in "sc" "wc"; do
+        cp "${CK8S_KUBESPRAY_PATH}"/config/common/group_vars/all/ck8s-kubespray-general.yaml \
+        "${CK8S_CONFIG_PATH}"/"${cluster}"-config/group_vars/all/ck8s-kubespray-general.yaml
+    done
+    ```
```
davidumea, line 24: This step should be added to more or another migration doc (`v2.21`?), as I don't think we will upgrade most clusters to this patch version, and the prepare scripts don't work if the file isn't created.

bin/common.bash
```diff
@@ -249,3 +249,21 @@ kubespray_version_check(){
...
+# Exits if they do not match.
+ck8s_kubespray_version_check(){
+    version=$(git describe --exact-match --tags 2> /dev/null || git rev-parse HEAD)
+    version_in_config=$(yq4 .ck8sKubesprayVersion "$config_path/group_vars/all/ck8s-kubespray-general.yaml")
+
+    if [[ -z "$version_in_config" || "$version_in_config" == "null" ]]; then
+        log_error "ERROR: No version set. Ensure that $config_path/group_vars/all/ck8s-kubespray-general.yaml exists and has ck8sKubesprayVersion set."
+        exit 1
+    elif [[ "${version_in_config}" != "any" ]] \
+        && if [[ "${version}" != "${version_in_config}" ]]; then
```
davidumea, line 263:
```suggestion
        && [[ "${version}" != "${version_in_config}" ]]; then
```

migration/v2.20.0-ck8s4/upgrade-cluster.md
Ajarmar: I was planning on migrating all environments to this patch version, don't need to do it during a maintenance window since the only difference is the addition of this file. But regardless of that I agree that this migration step should be added to the migration docs for `v2.21`, because we shouldn't have `v2.20.0-ck8s4` as a strictly required part of the upgrade path

migration/v2.21/prepare/12-ntp-servers.sh
```diff
@@ -0,0 +1,17 @@
...
+source "${ROOT}/scripts/migration/lib.sh"
+
+yq_add sc k8s_cluster/ck8s-k8s-cluster .ntp_enabled true
+yq_add sc k8s_cluster/ck8s-k8s-cluster .ntp_manage_config true
+yq_add sc k8s_cluster/ck8s-k8s-cluster .ntp_servers '[ "gbg1.ntp.netnod.se iburst", "gbg2.ntp.netnod.se iburst", "lul1.ntp.netnod.se iburst", "lul2.ntp.netnod.se iburst", "mmo1.ntp.netnod.se iburst", "mmo2.ntp.netnod.se iburst", "sth1.ntp.netnod.se iburst", "sth2.ntp.netnod.se iburst", "sth3.ntp.netnod.se iburst", "sth4.ntp.netnod.se iburst", "svl1.ntp.netnod.se iburst", "svl2.ntp.netnod.se iburst"]'
+yq_add sc k8s_cluster/ck8s-k8s-cluster .ntp_timezone "\"Etc/UTC\""
+
+yq_add wc k8s_cluster/ck8s-k8s-cluster .ntp_enabled true
+yq_add wc k8s_cluster/ck8s-k8s-cluster .ntp_manage_config true
+yq_add wc k8s_cluster/ck8s-k8s-cluster .ntp_servers '[ "gbg1.ntp.netnod.se iburst", "gbg2.ntp.netnod.se iburst", "lul1.ntp.netnod.se iburst", "lul2.ntp.netnod.se iburst", "mmo1.ntp.netnod.se iburst", "mmo2.ntp.netnod.se iburst", "sth1.ntp.netnod.se iburst", "sth2.ntp.netnod.se iburst", "sth3.ntp.netnod.se iburst", "sth4.ntp.netnod.se iburst", "svl1.ntp.netnod.se iburst", "svl2.ntp.netnod.se iburst"]'
```
davidumea, line 16: There should be something here (maybe a prompt) to ask if the environment you are upgrading is hosted in Sweden. Because if not the operator needs to manually add other appropriate ntp servers.

Ajarmar: Added a prompt

davidumea (approved):
Didn't try the new migration steps, but seems fine to me

davidumea:
I think I wrote it somewhere else too, but I'd say this change warrants a patch release

aarnq (approved):
Only comment I have is that perhaps we should collect migration/apply scripts under `apply/` as we do in apps to further clean up the structure. Else LGTM.

Ajarmar:
> I think I wrote it somewhere else too, but I'd say this change warrants a patch release

Yeah, I'll create patch releases for v2.20 and v2.21.

> Only comment I have is that perhaps we should collect migration/apply scripts under `apply/` as we do in apps to further clean up the structure. Else LGTM.

I like that idea, pushed a change.

bin/common.bash
```diff
@@ -250,3 +250,21 @@ kubespray_version_check(){
         fi
     fi
 }
+
+# Compares the checked out version and the config version of ck8s-kubespray.
+# Exits if they do not match.
+ck8s_kubespray_version_check(){
+    version=$(git describe --exact-match --tags 2> /dev/null || git rev-parse HEAD)
+    version_in_config=$(yq4 .ck8sKubesprayVersion "$config_path/group_vars/all/ck8s-kubespray-general.yaml")
```
OlleLarsson, line 258: Let's use `{}` in variable expansion to stay consistent :)
```suggestion
    version_in_config=$(yq4 .ck8sKubesprayVersion "${config_path}/group_vars/all/ck8s-kubespray-general.yaml")
```

bin/upgrade.bash
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
+
+here="$(readlink -f "$(dirname "${0}")")"
+
+ROOT="$(readlink -f "${here}/../")"
+
+CK8S_STACK="$(basename "$0")"
+export CK8S_STACK
```
OlleLarsson, line 8:
```suggestion
export CK8S_STACK="$(basename "$0")"
```

bin/upgrade.bash
```diff
@@ -0,0 +1,114 @@
...
+
+# shellcheck source=scripts/migration/lib.sh
+CK8S_ROOT_SCRIPT="true" source "${ROOT}/scripts/migration/lib.sh"
```
OlleLarsson, line 11:
```suggestion
CK8S_ROOT_SCRIPT="true"
source "${ROOT}/scripts/migration/lib.sh"
```

migration/README.md
```diff
@@ -4,11 +4,10 @@
 
 There is a script that can help you create the initial migration folder and `upgrade-cluster.md` document.
 
-This script must have ck8s-kubespray old and new versions as arguments including patch versions.
-If patch version does not matter for this migration then replace it with x. The second argument should include the full new version.
+This script must have ck8s-kubespray old and new versions as arguments. These arguments should include as much of the versions as is necessary; for new minor versions include major/minor version, for new patch versions include major/minor/patch version, etc.
```
OlleLarsson, line 7:
```suggestion
This script must have ck8s-kubespray old and new versions as arguments.
These arguments should include as much of the versions as is necessary; for new minor versions include major/minor version, for new patch versions include major/minor/patch version, etc.
```

bin/upgrade.bash
davidumea: Could be wrong but I think shellcheck complains if you do this

PR #276, opened 2023-06-12
WIP: Create managed upgrade script

### LINKED ISSUE #268: [5] Create managed upgrade script
To be safe the default upgrade method is to take all nodes in sequence in order to avoid situations were multiple nodes are upgraded which may cause downtime if it manages to pull multiple nodes per service.

But this takes ages, and we want to constrain the time an upgrade takes so it is more predictable both for operator and for users.
And we want to make this as managed as possible to reduce possible operator mistakes.

- I want to see a static inventory where we can define groups of different nodes that can be upgraded in parallel.
- I want to see a script that can run a playbook in a managed way that targets first the control plane (either only the first node since it is special in the eyes of Kubespray) and then the groups in parallel.
  - Of course it should only take one node per group in sequence!
- I also want to see the possibility of adding some type of check for the best node to upgrade, so if you have say database nodes using primary-secondary replication it would actually check which node to pick in order to move through the secondaries first.

**Definition of done:**
Upgrades are improved.

ayoubeddafali:

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #268 

**Checklist:**

  - [ ] will break the cluster.


ayoubeddafali:
Still a WIP, I just opened the PR to avoid losing the changes once again

bin/generate-groups-inventory.bash
```diff
@@ -0,0 +1,111 @@
...
+
+if [ -e "${config[groups_inventory_file]}" ]; then 
+    log_info "Using ${config[groups_inventory_file]} .."
+else 
+    log_info "Creating ${config[groups_inventory_file]} .."
+    cp "${config[inventory_file]}" "${config[groups_inventory_file]}"
+fi
+
+
+# Group Elastisys Nodes
```
robinAwallace, line 36: Would it be possible to make it find the groups automatically? 
So it finds control plane, elastitys etc.
For nodes with the label `elastisys.io/ams-cluster-name` you can do you cluster grouping.

So that we do not need to remember to add more specific groups here in the future.

Also i guess all nodes with out a label should be grouped up into a group?

ayoubeddafali: I honestly not sure If I get your point, I think the script will find the groups automatically, for example for elastisys nodes, it will check if there are any, if not we won't have any group of that sort in the new inventory ( same for AMS clusters .. ) 

Also, I can create a dedicated group for the controle plane, but not sure how the new inventory would affect the upgrade playbook, does kubespray expect a specific group name in the inventory for the controle plane  nodes ( ie: kube_controle_plane & kube_node ), currently I'm creating the new inventory based on the inventory.ini used to install kubespray, and later adding the groups for different type of nodes

robinAwallace: I mean that it also find the group `elastisys_node` automatically, etc. 
So that other specific groups that we or anyone else creates in the future is picked. 

It should be fine with a `controle_plane` group.

robinAwallace: Okay, I maybe misunderstood the task then. I thought it was going to generate a inventory file from the cluster and then do the upgrade. 

I would say look at adding labels/taints in the tfvars when creating new clusters.
I just not a fan of hard coding to our current labels.

aarnq: Yeah, now I haven't looked looked through the scripts to much but I do agree that we shouldn't hard code it, and that why I don't want it to look for our labels either.

Having a configurable list that takes labels+taints and a regex for node names should be enough to create a mapping from nodes to groups.

ayoubeddafali: @robinAwallace Actually I understand it as you described, the groups are generated from the actual environment

aarnq: I guess it's best if we skip my idea for now and focus on generating the inventory from the state of the cluster.

@ayoubeddafali You have hard coded the labels that you check on for groups and I would like that to configurable. Either like `group_label_primary: elastisys.io/node-type` and `group_label_secondary: elastisys.io/ams-cluster-name`, or in a list `group_labels: [elastisys.io/node-type, elastisys.io/ams-cluster-name]`. You can set this in the [new general file](https://github.com/elastisys/compliantkubernetes-kubespray/blob/main/config/common/group_vars/all/ck8s-kubespray-general.yaml)

bin/upgrade-groups.bash
```diff
@@ -0,0 +1,97 @@
...
+    for group in $(readInventoryGroups "${config[groups_inventory_file]}" | sort); do
+      if ! [[ "${group}" =~ ^(all|etcd|kube_node|.*:.*)$ ]]; then
+        groups+=("${group}")
+      fi
+    done
+
+    local -A group_lengths
+    group_lengths["kube_control_plane"]="$(getGroupHosts "${config[groups_inventory_file]}" "kube_control_plane" | wc -w)"
+    for group in "${groups[@]}"; do
+      group_lengths["${group}"]="$(getGroupHosts "${config[groups_inventory_file]}" "${group}" | wc -w)"
```
Xartos, line 48: Create a variable here to save the largest group so that you can use it for the index range later. Just to be sure we never exceed that number of nodes

```suggestion
      group_lengths["${group}"]="$(getGroupHosts "${config[groups_inventory_file]}" "${group}" | wc -w)"
      if [[ "${group_lengths["${group}"]}" -gt "${largest_group_length}" ]]; then
        largest_group_length="${group_lengths["${group}"]}"
      fi
```

bin/upgrade-groups.bash
```diff
@@ -0,0 +1,97 @@
...
+
+    popd
+}
+
+# TO DELETE
+get_group_logs() {
+    group=$1
+    shift
+    tail /tmp/$group.logs "${@}"
+}
```
Xartos, line 81:
```suggestion
```

bin/generate-groups-inventory.bash
```diff
@@ -0,0 +1,54 @@
...
+    if [[ "${reply}" != "y" ]]; then
+        exit 1
+    fi
+fi
+
+if [[ "$(groupExists ${config[inventory_file]} all)" == "true" ]]; then
+    all_section="$(getSection ${config[inventory_file]} all)"
+    echo -e "$all_section\n" > ${config[groups_inventory_file]}
+else 
+    log_error "Error: [all] group is defined in ${config[inventory_file]}"
```
Xartos, line 25: This should be not right?
```suggestion
    log_error "Error: [all] group is not defined in ${config[inventory_file]}"
```

Xartos (commented):
I assume that this was only made with a static inventory file in mind?
Because for openstack clusters we have a dynamic inventory file, [like this one](https://github.com/kubernetes-sigs/kubespray/blob/master/contrib/terraform/terraform.py). And it doesn't look like the parsing functions have that in mind. That would mean that only exoscale and upcloud is supported right now.

PR #309, opened 2023-08-31
Managed upgrade script

### LINKED ISSUE #268: [5] Create managed upgrade script

ayoub-eddafali:
### What kind of PR is this?

**Required**: Mark one of the following that is applicable:

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

*Optional*: Mark one or more of the following that are applicable:

> [!important]
> Critical security fixes should be marked with `kind/security`
> Breaking changes should be marked `kind/admin-change` or `kind/dev-change` depending on type

- [ ] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

### What does this PR do / why do we need this PR?

Managed upgrade script

- Fixes #268 

#### Checklist

- Change checks:
  - [] The change is transparent
  - [ ] The change is disruptive
  - [ ] The change requires no migration steps
  - [ ] The change requires migration steps
- Metrics checks:
  - [ ] The metrics are still exposed and present in Grafana after the change
  - [ ] The metrics names didn't change (Grafana dashboards and Prometheus alerts are not affected)
  - [ ] The metrics names did change (Grafana dashboards and Prometheus alerts were fixed)
- Logs checks:
  - [ ] The logs do not show any errors after the change
- Network Policy checks:
  - [ ] Any changed pod is covered by Network Policies
  - [ ] The change does not cause any dropped packages in the `NetworkPolicy Dashboard`
- Pod Security Policy checks:
  - [ ] Any changed pod is covered by Pod Security Admission
  - [ ] Any changed pod is covered by Gatekeeper Pod Security Policies
  - [ ] The change does not cause any pods to be blocked by Pod Security Admission or Policies
- Falco checks:
  - [ ] The change does not cause any alerts to be generated by Falco

### DIFF
--- a/bin/ck8s-kubespray
+++ b/bin/ck8s-kubespray
@@ -25,6 +25,12 @@ usage() {
     echo "      args: <wc|sc> [<options>]" 1>&2
     echo "  upgrade                                     prepares config for upgrade" 1>&2
     echo "      args: <wc|sc|both> <version> prepare" 1>&2
+    echo "  upgrade-groups                              upgrade groups in parallel" 1>&2
+    echo "      args: <wc|sc> <list-groups|apply>" 1>&2
+    echo "  generate-groups-inventory                   generate groups static inventory" 1>&2
+    echo "      args: <wc|sc>" 1>&2
+    echo "  sync-groups-inventory                       sync groups static inventory" 1>&2
+    echo "      args: <wc|sc>" 1>&2
     exit 1
 }
 
@@ -94,6 +100,27 @@ case "${1}" in
         fi
         shift 1
         "${here}/upgrade.bash" "${@}"
+        ;;
+    upgrade-groups)
+        if [ $# -lt 3 ]; then
+            usage
+        fi
+        shift 2
+        "${here}/upgrade-groups.bash" "${@}"
+        ;;
+    generate-groups-inventory)
+        if [ $# -lt 2 ]; then
+            usage
+        fi
+        shift 2
+        "${here}/generate-groups-inventory.bash" "${@}"
+        ;;
+    sync-groups-inventory)
+        if [ $# -lt 2 ]; then
+            usage
+        fi
+        shift 2
+        "${here}/sync-groups-inventory.bash" "${@}"
         ;;
     *) usage ;;
 esac
--- a/bin/common.bash
+++ b/bin/common.bash
@@ -41,6 +41,17 @@ export TERRAFORM_STATE_ROOT="${config_path}"
 declare -A config
 # shellcheck disable=SC2034
 config["inventory_file"]="${config_path}/inventory.ini"
+# shellcheck disable=SC2034
+config["groups_inventory_file"]="${config_path}/groups-inventory.ini"
+
+declare -A KUBE
+KUBE[sc]="${CK8S_CONFIG_PATH}/.state/kube_config_sc.yaml"
+KUBE[wc]="${CK8S_CONFIG_PATH}/.state/kube_config_wc.yaml"
+
+# shellcheck disable=SC2317
+log_info() {
+    echo -e "[\e[34mck8s\e[0m] ${*}" 1>&2
+}
 
 log_info_no_newline() {
     echo -e -n "[\e[34mck8s\e[0m] ${*}" 1>&2
@@ -332,4 +343,104 @@ ck8s_kubespray_version_check(){
         log_error "CK8S-Kubespray version: ${version}"
         exit 1
     fi
+}
+with_kubeconfig() {
+    kubeconfig="${1}"
+    shift
+
+    if [ ! -f "${kubeconfig}" ]; then
+      log_error "ERROR: Kubeconfig not found: ${kubeconfig}"
+      exit 1
+    fi
+
+    if grep -F -q 'sops:' "${kubeconfig}" || \
+        grep -F -q '"sops":' "${kubeconfig}" || \
+        grep -F -q '[sops]' "${kubeconfig}" || \
+        grep -F -q 'sops_version=' "${kubeconfig}"; then
+        log_info "Using encrypted kubeconfig ${kubeconfig}"
+
+        # TODO: Can't use a FIFO since we can't know that the kubeconfig is not
+        #       read multiple times. Let's try to eliminate the need for writing
+        #       the kubeconfig to disk in the future.
+        local -a args
+        for arg in "${@}"; do args+=("'${arg}'"); done
+        sops_exec_file_no_fifo "${kubeconfig}" "KUBECONFIG=\"{}\" ${args[*]}"
+    else
+        log_info "Using unencrypted kubeconfig ${kubeconfig}"
+        KUBECONFIG=${kubeconfig} "${@}"
+    fi
+}
+
+ops_kubectl() { # <prefix> <args...>
+    case "${1}" in
+        sc) kubeconfig="${KUBE[sc]}" ;;
+        wc) kubeconfig="${KUBE[wc]}" ;;
+    esac
+
+    shift
+    with_kubeconfig "$kubeconfig" kubectl "${@}"
+}
+
+assign_host() {
+    local node=$1
+    # Check for control plane nodes
+    control_plane_label=$(yq4 .control_plane_label "${config_path}/group_vars/all/ck8s-kubespray-general.yaml")
+    # Check for AMS
+    primary_group_label=$(yq4 .group_label_primary "${config_path}/group_vars/all/ck8s-kubespray-general.yaml")
+    secondary_group_label=$(yq4 .group_label_secondary "${config_path}/group_vars/all/ck8s-kubespray-general.yaml")
+    if [[ $(ops_kubectl "$prefix" get node "$node" -ojson | jq ".metadata.labels | has(\"${control_plane_label}\")") == "true" ]]; then
+        target_group="kube_control_plane"
+        if [[ "$(group_exists "${config[groups_inventory_file]}" "$target_group")" != "true" ]]; then
+            log_info "Adding $target_group group to ${config[groups_inventory_file]} .."
+            add_group "${config[groups_inventory_file]}" "$target_group"
+        fi
+        add_host_to_group "${config[groups_inventory_file]}" "$node" "$target_group"
+    elif [[ $(ops_kubectl "$prefix" get node "$node" -ojson | jq ".metadata.labels | has(\"${primary_group_label}\")") == "true" ]]; then
+        node_type=$(ops_kubectl "$prefix" get node "$node" -ojson | jq -r ".metadata.labels[\"${primary_group_label}\"]")
+        if [[ $(ops_kubectl "$prefix" get node "$node" -ojson | jq ".metadata.labels | has(\"${secondary_group_label}\")") == "true" ]]; then
+            cluster_name=$(ops_kubectl "$prefix" get node "$node" -ojson | jq -r ".metadata.labels[\"${secondary_group_label}\"]")
+            target_group="${node_type}_${cluster_name//-/_}"
+        else
+            target_group="${node_type}"
+        fi
+        if [[ "$(group_exists "${config[groups_inventory_file]}" "$target_group")" != "true" ]]; then
+            log_info "Adding $target_group group to ${config[groups_inventory_file]} .."
+            add_group "${config[groups_inventory_file]}" "$target_group"
+        fi
+
+        if [[ "$node_type" == "postgres" ]]; then
+            if [[ $(ops_kubectl "$prefix" get pods -A -l application=spilo,cluster-name="$cluster_name" -ojson | jq -r '.items[] | select( .metadata.labels["spilo-role"] == "master" ).spec.nodeName') == "$node" ]]; then
+                add_host_to_group_as_last "${config[groups_inventory_file]}" "$node" "$target_group"
+            else
+                add_host_to_group "${config[groups_inventory_file]}" "$node" "$target_group"
+            fi
+        elif [[ "$node_type" == "redis" ]]; then
+            if [[ $(ops_kubectl "$prefix" get pods -A -l redisfailovers.databases.spotahome.com/name="$cluster_name" -ojson | jq -r '.items[] | select( .metadata.labels["redisfailovers-role"] == "master" ).spec.nodeName') == "$node" ]]; then
+                add_host_to_group_as_last "${config[groups_inventory_file]}" "$node" "$target_group"
+            else
+                add_host_to_group "${config[groups_inventory_file]}" "$node" "$target_group"
+            fi
+        else
+            add_host_to_group "${config[groups_inventory_file]}" "$node" "$target_group"
+        fi
+
+
+    # Check for regular nodes
+    else
+        target_group="regular_worker"
+        if [[ "$(group_exists "${config[groups_inventory_file]}" "$target_group")" != "true" ]]; then
+            log_info "Adding $target_group group to ${config[groups_inventory_file]} .."
+            add_group "${config[groups_inventory_file]}" "$target_group"
+        fi
+        add_host_to_group "${config[groups_inventory_file]}" "$node" "$target_group"
+    fi
+}
+
+contains_element () {
+  local e match="$1"
+  shift
+  for e in "$@"; do
+    [[ "${e}" == "${match}" ]] && return 0;
+  done
+  return 1
 }
--- a/bin/generate-groups-inventory.bash
+++ b/bin/generate-groups-inventory.bash
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+
+here="$(dirname "$(readlink -f "$0")")"
+# shellcheck source=bin/common.bash
+source "${here}/common.bash"
+
+# shellcheck source=bin/common.bash
+source "${here}/inventory-parser.bash"
+
+# shellcheck disable=SC2154
+if [ -e "${config[groups_inventory_file]}" ]; then
+    log_info_no_newline "${config[groups_inventory_file]} will be overwritten, Proceed [y/N] ? "
+    read -n 1 -r reply
+    if [[ "${reply,,}" != "y" ]]; then
+        exit 1
+    fi
+fi
+
+if [[ "$(group_exists "${config[inventory_file]}" all)" == "true" ]]; then
+    all_section="$(get_section "${config[inventory_file]}" all)"
+    echo -e "$all_section\n" > "${config[groups_inventory_file]}"
+else
+    log_error "Error: [all] group is not defined in ${config[inventory_file]}"
+    exit 1
+fi
+
+if [[ "$(group_exists "${config[inventory_file]}" kube_control_plane)" == "true" ]]; then
+    kube_control_plane="$(get_section "${config[inventory_file]}" kube_control_plane)"
+    echo -e "$kube_control_plane\n" >> "${config[groups_inventory_file]}"
+else
+    log_error "Error: [kube_control_plane] group is not defined in ${config[inventory_file]}"
+    exit 1
+fi
+
+if [[ "$(group_exists "${config[inventory_file]}" etcd)" == "true" ]]; then
+    etcd_section="$(get_section "${config[inventory_file]}" etcd)"
+    echo -e "$etcd_section\n" >> "${config[groups_inventory_file]}"
+else
+    log_error "Error: [etcd] group is not defined in ${config[inventory_file]}"
+    exit 1
+fi
+
+if [[ "$(group_exists "${config[inventory_file]}" kube_node)" == "true" ]]; then
+    kube_node_section="$(get_section "${config[inventory_file]}" kube_node)"
+    echo -e "$kube_node_section\n" >> "${config[groups_inventory_file]}"
+else
+    log_error "Error: [kube_node] group is not defined in ${config[inventory_file]}"
+    exit 1
+fi
+
+# shellcheck disable=SC2091
+if [[ "$(group_exists "${config[inventory_file]}" k8s_cluster)" == "true" ]]; then
+  if $(group_has_children "${config[inventory_file]}" k8s_cluster); then
+    k8s_cluster_children="$(get_group_children "${config[inventory_file]}" k8s_cluster)"
+    add_group "${config[groups_inventory_file]}" "k8s_cluster:children"
+    echo -e "$k8s_cluster_children" >> "${config[groups_inventory_file]}"
+  else
+    k8s_cluster_section="$(get_section "${config[inventory_file]}" k8s_cluster)"
+    echo -e "$k8s_cluster_section\n" >> "${config[groups_inventory_file]}"
+  fi
+else
+    log_error "Error: [k8s_cluster:children] group is not defined in ${config[inventory_file]}"
+fi
+
+# shellcheck disable=SC2154
+nodes=$(ops_kubectl "$prefix" get nodes -o=jsonpath='{.items[*].metadata.name}')
+
+for node in $nodes; do
+    assign_host "$node"
+done
--- a/bin/inventory-parser.bash
+++ b/bin/inventory-parser.bash
@@ -1,0 +1,404 @@
+#!/bin/bash
+
+set -e -o pipefail
+
+here="$(dirname "$(readlink -f "$0")")"
+# shellcheck source=bin/common.bash
+source "${here}/common.bash"
+
+# Cache hosts info
+output_hosts_info(){
+  local filename="$1"
+  # shellcheck disable=SC2016
+  cluster_path=$(dirname "${filename}")
+  if [[ -f "${cluster_path}/terraform.tfstate" ]]; then
+    checksum=$(cat "${cluster_path}/terraform.tfstate" "${filename}" | md5sum | awk '{print $1}')
+  else
+    checksum=$(md5sum "${filename}" | awk '{print $1}')
+  fi
+  if ! [[ -f "/tmp/${checksum}.cache" ]]; then
+    ansible-inventory -i "${filename}" --list --output "/tmp/${checksum}.cache"
+  fi
+  cat "/tmp/${checksum}.cache"
+}
+
+# Get inventory groups
+# Args: <inventory>
+read_inventory_groups(){
+  local filename="$1"
+  output_hosts_info "${filename}" | jq -r ". | keys[] | select(. != \"_meta\")"
+}
+
+# Get the hosts of a specific group
+# Args: <inventory> <group>
+get_group_hosts() {
+  local filename="$1"
+  local section="$2"
+  hosts=()
+
+  # shellcheck disable=SC2091
+  if [[ ${section} == "all" ]]; then
+    readarray -td '' hosts < <(output_hosts_info "${filename}" |  jq -r "._meta.hostvars | keys[]")
+  elif $(output_hosts_info "${filename}"| jq ".$section | has(\"hosts\")"); then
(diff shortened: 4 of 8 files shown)


bin/common.bash
```diff
@@ -268,3 +274,95 @@ ck8s_kubespray_version_check(){
...
+    if [[ $(ops_kubectl $prefix get node "$node" -ojson | jq ".metadata.labels | has(\"${control_plane_label}\")") == "true" ]]; then
+        target_group="kube_control_plane"
+        if [[ "$(groupExists ${config[groups_inventory_file]} $target_group)" != "true" ]]; then
+            log_info "Adding $target_group group to ${config[groups_inventory_file]} .."
+            addGroup "${config[groups_inventory_file]}" "$target_group"
+        fi
+        addHostToGroup "${config[groups_inventory_file]}" "$node" "$target_group"
+    # Check for AMS
+    primary_group_label=$(yq4 .group_label_primary "${config_path}/group_vars/all/ck8s-kubespray-general.yaml")
+    secondary_group_label=$(yq4 .group_label_secondary "${config_path}/group_vars/all/ck8s-kubespray-general.yaml")
```
Xartos, line 327: I assume these variables should be here

```suggestion
    control_plane_label=$(yq4 .control_plane_label "${config_path}/group_vars/all/ck8s-kubespray-general.yaml")
    # Check for AMS
    primary_group_label=$(yq4 .group_label_primary "${config_path}/group_vars/all/ck8s-kubespray-general.yaml")
    secondary_group_label=$(yq4 .group_label_secondary "${config_path}/group_vars/all/ck8s-kubespray-general.yaml")
    if [[ $(ops_kubectl $prefix get node "$node" -ojson | jq ".metadata.labels | has(\"${control_plane_label}\")") == "true" ]]; then
        target_group="kube_control_plane"
        if [[ "$(groupExists ${config[groups_inventory_file]} $target_group)" != "true" ]]; then
            log_info "Adding $target_group group to ${config[groups_inventory_file]} .."
            addGroup "${config[groups_inventory_file]}" "$target_group"
        fi
        addHostToGroup "${config[groups_inventory_file]}" "$node" "$target_group"

```

bin/generate-groups-inventory.bash
```diff
@@ -0,0 +1,54 @@
...
+# shellcheck source=bin/common.bash
+source "${here}/common.bash"
+
+# shellcheck source=bin/common.bash
+source "${here}/inventory-parser.bash"
+
+if [ -e "${config[groups_inventory_file]}" ]; then 
+    log_info_no_newline "${config[groups_inventory_file]} will be overwritten, Proceed [y/N] ? "
+    read -r reply
+    if [[ "${reply}" != "y" ]]; then
```
Xartos, line 16: Just to also allow capital `y`
```suggestion
    if [[ ! "${reply}" =~ ^[yY]$ ]]; then
```

bin/upgrade-groups.bash
```diff
@@ -0,0 +1,88 @@
...
+    skip_tags="multus"
+    if [[ "${1}" == "--skip-download" ]]; then
+      skip_tags+=",download"
+    fi
+    ansible-playbook upgrade-cluster.yml -b -i "${config[groups_inventory_file]}" --skip-tags=${skip_tags} --limit "kube_control_plane[0]"
+
+    for index in $(seq 0 100); do
+      local -a limit
+      limit=()
+      for group in "${groups[@]}"; do
```
Xartos, line 59: Wouldn't you be able to swap these two for loops? Then you could use the `${group_lengths["${group}"]}` to set the upper limit for the `index` variable. Just to reduce "magic numbers"

bin/upgrade-groups.bash
```diff
@@ -0,0 +1,88 @@
...
+            limit+=("${group}[$((index+1))]")
+          fi
+        elif [[ "${index}" -lt "${group_lengths["${group}"]}" ]]; then
+          limit+=("${group}[${index}]")
+        fi
+      done
+      if [[ -z "${limit[*]}" ]]; then
+        break
+      fi
+      ansible-playbook upgrade-cluster.yml -uroot -b -i "${config[groups_inventory_file]}" --skip-tags=${skip_tags} --limit "$(tr ' ' ',' <<< "${limit[*]}")" -e serial=100%
```
Xartos, line 71: Maybe it's too big to include in this issue, but it would be super nice if we could have a script where it ran each node in the group one by one and between them it executes something that you give it until it executes successfully.
This would be great for the rook-ceph group, since we need to wait for the ceph cluster to be `HEALTH_OK` before proceeding with the next node when upgrading them. Otherwise we risk corrupting/loose data.

But if you feel that this is too big or out of scope for this issue then we can create a separate issue for that.

bin/upgrade-groups.bash
```diff
@@ -0,0 +1,88 @@
...
+    for group in "${groups[@]}"; do
+      group_lengths["${group}"]="$(getGroupHosts "${config[groups_inventory_file]}" "${group}" | wc -w)"
+    done
+
+    pushd "${kubespray_path}"
+    skip_tags="multus"
+    if [[ "${1}" == "--skip-download" ]]; then
+      skip_tags+=",download"
+    fi
+    ansible-playbook upgrade-cluster.yml -b -i "${config[groups_inventory_file]}" --skip-tags=${skip_tags} --limit "kube_control_plane[0]"
```
Xartos, line 54: Why do we want to upgrade only the first control plane node? Shouldn't this be

```suggestion
    ansible-playbook upgrade-cluster.yml -b -i "${config[groups_inventory_file]}" --skip-tags=${skip_tags} --limit "kube_control_plane"
```

bin/upgrade-groups.bash
Xartos, line 71: This should run in the background or in a separate terminal. Since doing the groups in series doesn't make any sense. Then we could've ran it without the `--limit` flag.

aarnq:
Would it be possible to use `ansible-inventory -i inventory.ini --list` instead to get all required information from the main inventory? It would be able to handle generated inventories as we have with OpenStack environments, and also make it easier to parse and manage as it outputs in JSON so you can easily list groups and hosts.

bin/upgrade-groups.bash
aarnq: This is so it will upgrade the first control-plane node before it runs the rest in parallel with other nodes.

bin/upgrade-groups.bash
aarnq: This runs one node from each group at the same time, `serial=100%` means run in batches of 100% of the hosts, so all that the limit matches.
And I would refrain from running anything in the background yet as that give way less observability and I think this is quite important when we change the process this much.

bin/upgrade-groups.bash
aarnq: Not really, since the top for loop needs to run for the maximum number of nodes in one of of the group, and not the number of groups, as it build the next batch of nodes to be upgraded by picking the next node of each group.

aarnq:
Nit: Do try to fix the pre-commit issues, and usually with the code style we use function names should be `snake_case`.

bin/common.bash
```diff
@@ -268,3 +274,95 @@ ck8s_kubespray_version_check(){
...
+
+    if [ ! -f "${kubeconfig}" ]; then
+      log_error "ERROR: Kubeconfig not found: ${kubeconfig}"
+      exit 1
+    fi
+
+    if grep -F -q 'sops:' "${kubeconfig}" || \
+        grep -F -q '"sops":' "${kubeconfig}" || \
+        grep -F -q '[sops]' "${kubeconfig}" || \
+        grep -F -q 'sops_version=' "${kubeconfig}"; then
```
Zash, line 289: Are all these variants needed?

bin/generate-groups-inventory.bash
```diff
@@ -0,0 +1,54 @@
...
+here="$(dirname "$(readlink -f "$0")")"
+# shellcheck source=bin/common.bash
+source "${here}/common.bash"
+
+# shellcheck source=bin/common.bash
+source "${here}/inventory-parser.bash"
+
+if [ -e "${config[groups_inventory_file]}" ]; then 
+    log_info_no_newline "${config[groups_inventory_file]} will be overwritten, Proceed [y/N] ? "
+    read -r reply
```
Zash, line 15:
```suggestion
    read -n 1 -r reply
```
> `help read`
> -n nchars	return after reading NCHARS characters rather than waiting for a newline, but honor a delimiter if fewer than

so there's no chance of anyone typing the full 'yes' and you don't have to press Enter

bin/generate-groups-inventory.bash
Xartos: TIL :)

bin/common.bash
ayoubeddafali: Yes, this method was actually taken from the apps repo ( https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/common.bash#L552 )

ayoubeddafali: Seems like it has been updated there, so I'll probably refresh it here as well

aarnq:
> Would it be possible to use `ansible-inventory -i inventory.ini --list` instead to get all required information from the main inventory? It would be able to handle generated inventories as we have with OpenStack environments, and also make it easier to parse and manage as it outputs in JSON so you can easily list groups and hosts.

I still think this should be done within this PR to make it work with generated inventories.

bin/ck8s-kubespray
```diff
@@ -25,6 +25,11 @@ usage() {
     echo "      args: <prefix> [<options>]" 1>&2
     echo "  upgrade                                     prepares config for upgrade" 1>&2
     echo "      args: <wc|sc|both> <version> prepare" 1>&2
+    echo "      args: <version> prepare" 1>&2
```
anders-elastisys, line 28: Is there a reason why this line was added?

ayoubeddafali: Not really, I did a rebase and that was the source of the conflict, so I kept it to not override any changes.
I guess as you've worked on splitting the upgrade, I suppose that line is no longer needed, right ?

anders-elastisys: That makes sense, yes it should be fine to remove :slightly_smiling_face:

anders-elastisys (commented):
Now that [this PR](https://github.com/elastisys/compliantkubernetes-kubespray/pull/330) has been merged to main, could you rebase and change `<prefix>` occurances in _usage_ messages to `<wc|sc>`?

bin/common.bash
```diff
@@ -41,6 +41,17 @@ export TERRAFORM_STATE_ROOT="${config_path}"
 declare -A config
 # shellcheck disable=SC2034
 config["inventory_file"]="${config_path}/inventory.ini"
+# shellcheck disable=SC2034
+config["groups_inventory_file"]="${config_path}/groups-inventory.ini"
+
+declare -A KUBE
+KUBE[sc]="${CK8S_CONFIG_PATH:-}/.state/kube_config_sc.yaml"
+KUBE[wc]="${CK8S_CONFIG_PATH:-}/.state/kube_config_wc.yaml"
```
Xartos, line 49: NIT

```suggestion
KUBE[sc]="${CK8S_CONFIG_PATH}/.state/kube_config_sc.yaml"
KUBE[wc]="${CK8S_CONFIG_PATH}/.state/kube_config_wc.yaml"
```

We can be sure that this is already set from earlier in the file

bin/common.bash
```diff
@@ -333,3 +344,95 @@ ck8s_kubespray_version_check(){
...
+            add_group "${config[groups_inventory_file]}" "$target_group"
+        fi
+        add_host_to_group "${config[groups_inventory_file]}" "$node" "$target_group"
+    fi
+}
+
+contains_element () {
+  local e match="$1"
+  shift
+  for e; do [[ "$e" == "$match" ]] && return 0; done
```
Xartos, line 436: How does this work exactly? `e` doesn't seem to be assigned anywhere. And still you can iterate over it and use the same name as the element? 
It would be nice, imo, if you could expand this a little bit just to make it a bit more readable

bin/generate-groups-inventory.bash
```diff
@@ -0,0 +1,62 @@
...
+    if [[ "${reply,,}" != "y" ]]; then
+        exit 1
+    fi
+fi
+
+if [[ "$(group_exists "${config[inventory_file]}" all)" == "true" ]]; then
+    all_section="$(get_section "${config[inventory_file]}" all)"
+    echo -e "$all_section\n" > "${config[groups_inventory_file]}"
+else
+    log_error "Error: [all] group is not defined in ${config[inventory_file]}"
```
Xartos, line 26: Is this fatal enough to need to exit?

```suggestion
    log_error "Error: [all] group is not defined in ${config[inventory_file]}"
    exit 1
```

ayoubeddafali: Not sure to be honest, but I suppose kubespray relies on all to be defined for its playbooks

bin/common.bash
ayoubeddafali: Yep, I agree, that might be confusing, as the goal is to check if an element exist in an array.

**shift** shifts the argument list by 1 to the left (dropping the first argument) and for without an in implicitly iterates over the argument list

Xartos: Right, it was that `for` magic with out using `in` that I didn't know about. But maybe just expand it to

```bash
for e in "$@"; do
  [[ "$e" == "$match" ]] && return 0
done
```

Would make it way more readable in my book.

ayoubeddafali: Sure, will change that. Thanks a lot

bin/generate-groups-inventory.bash
Xartos: So as it is right now it isn't exiting. So my question was (I see that my earlier question was confusing) if you think we should exit instead of leaving as it is right now and not exit.
But then if I hear you correctly you think that we should exit

bin/upgrade-groups.bash
ayoubeddafali: After some attempts, I honestly couldn't make this work, as things kept breaking and bash gave me a hard time, so I removed that part from the implementation, and will create a separate ticket for it

ayoubeddafali:
Added some fixes, and tested it again on an empty exoscale kubespray cluster, and seems to be working on.
Would be grateful if someone could test this on his dev cluster, or a cluster with AMSes

aarnq (commented):
Worked well on a cluster with AMSes. I did some modification to be able to run it, and to run the control-plane serially first.

When it came to the groups it managed to do it in three batches with 5 / 5 / 4 nodes and with download it took 31 min in total (including the time for the control-plane).

bin/inventory-parser.bash
```diff
@@ -0,0 +1,378 @@
...
+
+here="$(dirname "$(readlink -f "$0")")"
+# shellcheck source=bin/common.bash
+source "${here}/common.bash"
+
+# Get inventory groups
+# Args: <inventory>
+read_inventory_groups(){
+  local filename="$1"
+  ansible-inventory -i "$1" --list | jq -r ". | keys[] | select(. != \"_meta\")"
```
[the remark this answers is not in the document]
ayoubeddafali: Yes you are right.  I tried a straightforward implementation in caching the response, seems like it works fine ( went from 2minutes to 20 seconds for my dev cluster). 
Would be grateful if you could regenerate the groups inventory, and see if nothing breaks

bin/ck8s-kubespray
```diff
@@ -25,6 +25,10 @@ usage() {
     echo "      args: <wc|sc> [<options>]" 1>&2
     echo "  upgrade                                     prepares config for upgrade" 1>&2
     echo "      args: <wc|sc|both> <version> prepare" 1>&2
+    echo "  generate-groups-inventory                   generate groups static inventory" 1>&2
+    echo "      args: <wc|sc>" 1>&2
+    echo "  sync-groups-inventory                       sync groups static inventory" 1>&2
+    echo "      args: <wc|sc>" 1>&2
```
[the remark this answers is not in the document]
ayoubeddafali: Added 👍🏽

bin/upgrade-groups.bash
```diff
@@ -0,0 +1,88 @@
...
+
+# shellcheck source=bin/common.bash
+source "${here}/inventory-parser.bash"
+
+list_groups() {
+    groups=$(read_inventory_groups "${config[groups_inventory_file]}")
+    for group in $groups; do
+        log_info "- $group"
+    done
+}
```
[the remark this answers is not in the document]
ayoubeddafali: Added as well 👍🏽

bin/upgrade-groups.bash
```diff
@@ -0,0 +1,88 @@
...
+            limit+=("${group}[$((index+1))]")
+          fi
+        elif [[ "${index}" -lt "${group_lengths["${group}"]}" ]]; then
+          limit+=("${group}[${index}]")
+        fi
+      done
+      if [[ -z "${limit[*]}" ]]; then
+        break
+      fi
+      ansible-playbook upgrade-cluster.yml -uroot -b -i "${config[groups_inventory_file]}" --skip-tags=${skip_tags} --limit "$(tr ' ' ',' <<< "${limit[*]}")" -e serial=100%
```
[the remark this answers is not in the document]
ayoubeddafali: Added to check for the hostvar, if not found it will default to *ubuntu*

docs/kubespray-upgrade-groups.md
```diff
@@ -0,0 +1,44 @@
...
+
+The managed upgrade scripts should allow us to :
+
+- Generate a static inventory with different groups of nodes based on their type ( elastisys, worker, postgres, control plane ... etc)
+
+- Upgrade the control planes first
+
+- Upgrade the groups in parallel by taking only one node per group, in sequence
+
+- Pick the order in which nodes to upgrade first to avoid downtime for databases. In case of primary-secondary replication, we should start with node that host the secondaries first
```
aarnq, line 14:
```suggestion
- Pick the order in which nodes to upgrade first to avoid downtime for databases. In case of primary-secondary replication, we should start with node that host the secondaries first (not yet implemented).
```
Just to make it clear.

aarnq (commented):
I'll do another test.

bin/inventory-parser.bash
```diff
@@ -0,0 +1,390 @@
...
+
+here="$(dirname "$(readlink -f "$0")")"
+# shellcheck source=bin/common.bash
+source "${here}/common.bash"
+
+# Cache hosts info
+output_hosts_info(){
+  local filename="$1"
+  # shellcheck disable=SC2016
+  md5=$(echo -n "${filename}"| xargs -n1 -I{} bash -c 'echo "$1"; cat "$1"' x {} | md5sum | awk '{print $1}')
```
aarnq, line 13: I'm getting warnings form xargs for this line, what does it actually do?

docs/kubespray-upgrade-groups.md
ayoubeddafali: Actually, the script is getting the nodes which hosts the postgres pods that have the label *spilo-role* equal to master, and append them as the last element in the group, so they will be the last nodes to get upgraded

bin/inventory-parser.bash
ayoubeddafali: What kind of warning? as I didn't get any on my local machine. 
Regarding that line, the idea at first was getting the md5 of the inventory, but it seems like for dynamic ones it doesn't work, as we will be getting the same md5 for both sc/wc inventories, so I was trying to find a way to include both content + filename in the resulting md5

bin/inventory-parser.bash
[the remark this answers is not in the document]
aarnq: The current implementation will have some issues with dynamic inventories as the inventory.ini doesn't change the same way a static does, but that can be amended later. It does provide a _much_ better experience over no caching.

docs/kubespray-upgrade-groups.md
aarnq: Wha :sweat_smile: I missed that. Is it possible to make that configurable with a list of key=value pairs to look for?

bin/inventory-parser.bash
aarnq: That `-I` and `-n` are mutually exclusive so `-n` is skipped.

For dynamic inventories you can include `terraform.tfstate` in the checksum if it exists.

bin/upgrade-groups.bash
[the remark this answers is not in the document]
aarnq: It's not sufficient to set it when we running the upgrade, this should propagate from the main inventory to the group inventory.

docs/kubespray-upgrade-groups.md
aarnq: Where do you do this check?

bin/common.bash
```diff
@@ -333,3 +344,97 @@ ck8s_kubespray_version_check(){
...
+        else
+            target_group="${node_type}"
+        fi
+        if [[ "$(group_exists "${config[groups_inventory_file]}" "$target_group")" != "true" ]]; then
+            log_info "Adding $target_group group to ${config[groups_inventory_file]} .."
+            add_group "${config[groups_inventory_file]}" "$target_group"
+        fi
+
+        if [[ "$node_type" == "postgres" ]]; then
+            if [[ $(ops_kubectl "$prefix" get pods -A -l application=spilo,cluster-name="$cluster_name" -ojson | jq -r '.items[] | select( .metadata.labels["spilo-role"] == "master" ).spec.nodeName') == "$node" ]]; then
```
ayoubeddafali, line 412: @aarnq here

bin/upgrade-groups.bash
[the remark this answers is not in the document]
ayoubeddafali: hmm, don't you have the *ansible_user* variable set on the hosts in your generated groups-inventory ?

bin/common.bash
aarnq: Ah, thanks :+1:

bin/inventory-parser.bash
ayoubeddafali: Sure, I'll change it to get the checksum of the terraform.tfstate instead, that should be straightforward

bin/upgrade-groups.bash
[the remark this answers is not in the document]
aarnq: Ah, the dynamic inventory only output `ansible_ssh_user` which is a legacy name for that variable, so if you can propagate that one too then it should be fine.

bin/upgrade-groups.bash
```diff
@@ -0,0 +1,85 @@
...
+    for group in "${groups[@]}"; do
+      group_lengths["${group}"]="$(get_group_hosts "${config[groups_inventory_file]}" "${group}" | wc -w)"
+    done
+
+    pushd "${kubespray_path}"
+    skip_tags="multus"
+    if [[ "${1}" == "--skip-download" ]]; then
+      skip_tags+=",download"
+    fi
+    ansible-playbook upgrade-cluster.yml -b -i "${config[groups_inventory_file]}" --skip-tags=${skip_tags} --limit "kube_control_plane" --serial 1
```
aarnq, line 57:
```suggestion
    ansible-playbook upgrade-cluster.yml -b -i "${config[groups_inventory_file]}" --skip-tags=${skip_tags} --limit "kube_control_plane" -e serial=1
```

aarnq: I edited it :sweat_smile:

bin/sync-groups-inventory.bash
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+
+here="$(dirname "$(readlink -f "$0")")"
```
aarnq, line 6: This is broken for me, and I'm thinking that given how well generate groups inventory works this feels redundant.

ayoubeddafali: It should work for a static inventory, I'll try it on a dynamic one

bin/upgrade-groups.bash
ayoubeddafali: I've removed --user argument , as the variable should be present now on the groups-inventory

bin/upgrade-groups.bash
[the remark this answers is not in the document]
ayoubeddafali: It is added now

bin/sync-groups-inventory.bash
aarnq: The sync now works except for the `ansible_user` becomes null with `ansible_ssh_user`.

ayoubeddafali: Fixed the issues for the dynamic inventory, hopefully, it should work fine this time

docs/kubespray-upgrade-groups.md
ayoubeddafali: I think I didn't understand correctly the key=value pair configuration, is it specific to postgres ( ie: spilo-role=master )?

aarnq: I was thinking to make it more generic, so you can define for example `[spilo-role: master, redisfailovers-role: master]` in the config and it would check that if any node contains a pod matching any of the set labels it would just put that last in the group.

ayoubeddafali: I don't think it is feasible after some trial, as we're not relying only on those labels. It seems like for each component we need specific labels to determine the nodes of a cluster and then get the ones hosting the masters. 
I've added anyway support for `redisfailovers-role ` in a separate condition, so now nodes hosting redis masters will be the last to get upgraded

aarnq: Not really, we only need to do an early get on all namespaces with the configured labels and get the list of nodes that host those pods, then when we're adding to the inventory we just do a lookup and see if it exists in that list, then we should put it last in it's group.

I'm fine with doing that in a separate PR as this works although it is static.

bin/inventory-parser.bash
```diff
@@ -0,0 +1,401 @@
...
+# shellcheck source=bin/common.bash
+source "${here}/common.bash"
+
+# Cache hosts info
+output_hosts_info(){
+  local filename="$1"
+  # shellcheck disable=SC2016
+  path=$(dirname "${filename}")
+  # md5=$(echo -n "${filename}"| xargs -n1 -I{} bash -c 'echo "$1"; cat "$1"' x {}get_group_hosts | md5sum | awk '{print $1}')
+  md5=$(cat "${path}/terraform.tfstate" "${filename}" | md5sum | awk '{print $1}')
```
aarnq, line 15: Remove the comment and check if the state file exists before including it in the checksum.

ayoubeddafali: Added, thanks.

aarnq (approved):
LGTM, squash your commits before merging.

OlleLarsson (approved):

PR #344, opened 2024-01-16
docs: use skip_downloads flag instead of skip tags

### LINKED ISSUE #337: [3] Upgrade fails on etcdutl step
**Describe the bug**
The following step fails when running the `upgrade_cluster` playbook if the flag `--skip-tags=download` is used: https://github.com/kubernetes-sigs/kubespray/blob/v2.23.1/roles/etcdctl_etcdutl/tasks/main.yml#L23

The step is only run for master nodes.

**To Reproduce**

1. Run the kubespray upgrade for `v2.23` by following the guide.
2. When running `./bin/ck8s-kubespray run-playbook <sc|wc> upgrade_cluster.yml -b --skip-tags=download`, the user will receive an error on the master nodes:
  ```
  TASK [etcdctl_etcdutl : Copy etcd binary] **********************************************************
fatal: [<master-node-name>]: FAILED! => {"changed": false, "msg": "Source '/tmp/releases/etcd-v3.5.7-linux-amd64.tar.gz' does not exist"}
  ```

**Expected behavior**
No error should occur. It should be possible to run the download step and the rest of the upgrade separately.

- Compliant kubernetes kubespray v2.23.0-ck8s1 (when upgrading to this version)

**Additional context**
The reason why this fails is because, when using `etcd_deployment_type: kubeadm`, the file that the step is attempting to copy is not downloaded onto the node when running the `upgrade_cluster` playbook with the `--tags=download` flag separately (which we recommend in the migration guide).
The `etcdctl_etcdutl` role has a step to run the `download_file` role to download the file it tries to copy, but when running with `--skip-tags=download` this step is skipped - so the file will never be downloaded.

I have identified the following potential fixes to this:

- Use `-e skip_downloads=true` instead of `--skip-tags=download`. This should allow the `etcdctl_etcdutl` role to download the file, while skipping the "main" download. However, using this variable also causes other things to be skipped, such as `kubespray_defaults` and `kubernetes_preinstall`, so this must be tested to see that it works as expected. See this file: https://github.com/kubernetes-sigs/kubespray/blob/v2.23.1/playbooks/upgrade_cluster.yml#L26
- Make a change in kubespray to make sure the relevant file is downloaded even when using `etcd_deployment_type: kubeadm`.
- Use `etcd_deployment_type: host` instead of `kubeadm`. (I assume we use `kubeadm` for a reason, though.)

anders-elastisys:
### What kind of PR is this?

**Required**: Mark one of the following that is applicable:

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

*Optional*: Mark one or more of the following that are applicable:

> [!important]
> Critical security fixes should be marked with `kind/security`
> Breaking changes should be marked `kind/admin-change` or `kind/dev-change` depending on type

- [ ] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

### What does this PR do / why do we need this PR?

Change flags used in docs for when performing the disruptive part of a cluster upgrade, see issue for context.

- Fixes #337 

#### Checklist

- Change checks:
  - [x] The change is transparent
  - [ ] The change is disruptive
  - [ ] The change requires no migration steps
  - [ ] The change requires migration steps
- Metrics checks:
  - [ ] The metrics are still exposed and present in Grafana after the change
  - [ ] The metrics names didn't change (Grafana dashboards and Prometheus alerts are not affected)
  - [ ] The metrics names did change (Grafana dashboards and Prometheus alerts were fixed)
- Logs checks:
  - [ ] The logs do not show any errors after the change
- Network Policy checks:
  - [ ] Any changed pod is covered by Network Policies
  - [ ] The change does not cause any dropped packages in the `NetworkPolicy Dashboard`
- Pod Security Policy checks:
  - [ ] Any changed pod is covered by Pod Security Admission
  - [ ] Any changed pod is covered by Gatekeeper Pod Security Policies
  - [ ] The change does not cause any pods to be blocked by Pod Security Admission or Policies
- Falco checks:
  - [ ] The change does not cause any alerts to be generated by Falco

### DIFF
--- a/migration/template/upgrade-cluster.md
+++ b/migration/template/upgrade-cluster.md
@@ -46,8 +46,8 @@ These steps will cause disruptions in the environment.
 1. Upgrade the cluster to a new kubernetes version:
 
     ```bash
-    ./bin/ck8s-kubespray run-playbook sc upgrade_cluster.yml -b --skip-tags=download
-    ./bin/ck8s-kubespray run-playbook wc upgrade_cluster.yml -b --skip-tags=download
+    ./bin/ck8s-kubespray run-playbook sc upgrade_cluster.yml -b -e skip_downloads=true
+    ./bin/ck8s-kubespray run-playbook wc upgrade_cluster.yml -b -e skip_downloads=true
     ```
 
 ## Postrequisite
--- a/migration/v2.23/upgrade-cluster.md
+++ b/migration/v2.23/upgrade-cluster.md
@@ -47,8 +47,8 @@ These steps will cause disruptions in the environment.
 1. Upgrade the cluster to a new kubernetes version:
 
     ```bash
-    ./bin/ck8s-kubespray run-playbook sc upgrade_cluster.yml -b --skip-tags=download
-    ./bin/ck8s-kubespray run-playbook wc upgrade_cluster.yml -b --skip-tags=download
+    ./bin/ck8s-kubespray run-playbook sc upgrade_cluster.yml -b -e skip_downloads=true
+    ./bin/ck8s-kubespray run-playbook wc upgrade_cluster.yml -b -e skip_downloads=true
     ```
 
 ## Postrequisite
--- a/migration/v2.24/upgrade-cluster.md
+++ b/migration/v2.24/upgrade-cluster.md
@@ -46,8 +46,8 @@ These steps will cause disruptions in the environment.
 1. Upgrade the cluster to a new kubernetes version:
 
     ```bash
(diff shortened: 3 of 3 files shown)


anders-elastisys:
Let me know if I should also update older migration documents with this change.

Pavan-Gunda:
> Let me know if I should also update older migration documents with this change.

We can, also update the v23 migration document since that version has this etcdutl bug.

lunkan93 (approved):

